PR #8, opened 2023-02-17
Features/202302 tl module

sagar87:
* [first implementation of tl.label_property](https://github.com/scverse/spatialdata-plot/commit/25ab44215b7eca713229d3845d18f62740f9d4d1)
* [pp.get_bb also subsets the table](https://github.com/scverse/spatialdata-plot/commit/c22b6032b57e71384234d4e8d8815ddfef33430f)
* [pp.get_bb saves current bounding box boundaries in tables.uns](https://github.com/scverse/spatialdata-plot/commit/6054e018b4fb76799de5f28248a1700411f8b07c)

### DIFF
--- a/src/spatialdata_plot/pl/basic.py
+++ b/src/spatialdata_plot/pl/basic.py
@@ -83,7 +83,8 @@ class PlotAccessor:
         if num_images == 1:
             ax = ax or plt.gca()
             key = [k for k in image_data.keys()][0]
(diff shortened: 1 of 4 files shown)


timtreis:
@sagar87 I guess this PR is soo far away from the current main, we might as well close it? Or do we want to refactor it with the new version? I really like the functionality but don't have the capacity right now. If you're also busy right now, we could turn it into an "enhancement issue".

PR #17, opened 2023-03-09
fix docs

giovp:
- [x] fix build docs
- [x] add rtd CI
- [x] add pre-commit CI


giovp:
@timtreis @sagar87 I will merge this if you approve, even if tests don't pass cause not sure how to fix them (this PR is just to do things described above), but few additional comments:
- noticed you have "sandbox" notebooks in the root of the projects. I would remove them and add them in spatial sandbox repo or even in others but not here, diffs for notebooks, especially with many images, can get very heavy and then it's a pain to remove them to git history. 
- notebooks here should just be the ones for the docs, and actually we might not even want that, but we might want to just put them in `spatialdata-notebooks` repo. For now the simple one I would live but would not modify extensively until API is ready
- now there is CI for docs, meaning you can check how the docs build in URL (click details next to the job). You can also build them locally with `make html` in `docs` folder, but anyway useful to check out how they render.
- we might want to have branch protection in main, please feel free to add it following how it's done in other spatialdata repos
- I think we want `mypy` checks as well. Adding it now will be problematic (probably many fixes to be added), but adding it later also potentially. more difficult. It's gonna be quite some work, so you decide when you think is best.

timtreis (approved):

giovp:
will merge regardless of pre-commit, @timtreis @sagar87 you could fix in separate PR?

timtreis:
Yeah, I will fix this next

PR #18, opened 2023-03-13
mypy fixes

timtreis:


.pre-commit-config.yaml
```diff
@@ -5,14 +5,9 @@ default_stages:
...
+    skip: [mypy]
 repos:
-    - repo: https://github.com/nbQA-dev/nbQA
-      rev: 1.6.3
-      hooks:
-          - id: nbqa-pyupgrade
-            args: [--py38-plus]
-          - id: nbqa-black
-          - id: nbqa-isort
     - repo: https://github.com/psf/black
```
giovp, line 11: would not remove this cause we will have notebooks here I think even if just few

.pre-commit-config.yaml
```diff
@@ -29,6 +24,12 @@ repos:
       rev: 5.12.0
       hooks:
           - id: isort
+    - repo: https://github.com/pre-commit/mirrors-mypy
+      rev: v1.0.1
+      hooks:
+          - id: mypy
+            additional_dependencies: [numpy==1.22.0, types-requests]
+            exclude: tests/|docs/|temp/|spatialdata/_core/reader.py
```
giovp, line 32: is this the altest version? I would just run `pre-commit autoupdate` to automatically update all versions of pre-commit (it might be, just to be sure)

giovp (approved):
looks good! just two minor comments, then LGTM!

.pre-commit-config.yaml
timtreis: image

PR #26, opened 2023-03-30
Added tests to render_images

timtreis:


src/spatialdata_plot/pl/render.py
```diff
@@ -116,6 +111,7 @@ def _render_images(
             pca.fit(flattened_img.T)
             transformed_image = pca.transform(flattened_img.T)
             img = xr.DataArray(transformed_image.T.reshape(3, y_dim, x_dim), dims=("c", "y", "x"))
+            colors = ListedColormap(["red", "blue", "green"])
 
     img = xr.DataArray(img, dims=("c", "y", "x")).transpose("y", "x", "c")  # for plotting
```
sagar87, line 116: Better use just `numpy` to do the transpose, something like `img.T.swapaxes()` ? No need to cast to a `xarray` dataarray just for the sake of transpose.

sagar87 (approved):
Could do the `numpy` transpose instead of casting to `xarray` and then transposing in `src/spatialdata_plot/pl/render.py`.

PR #40, opened 2023-04-04
Basic structure for comparing generated plots in unit tests

giovp:
@timtreis I would potentially merge this as is and once points and polygons are working than we can start adding more tests to cover ground.


giovp:
tests are now failing cause the `blobs` PR is not in main yet.

timtreis:
Looks good, waiting for aggregate PR to be merged to merge this

PR #45, opened 2023-04-05
add squidpy way to handle segment

giovp:


giovp:
@sagar87 @timtreis this is a heavy rewrite of the rendering functions. It mostly focuses on aesthetics (colors, legend, colorbars, etc.) and doesn't change the core behavior of the library.

With this current implementation I can showcase the aggregation functions in the notebooks but I'm afraid is not enough for the datasets you have been working on. I would however suggest to re-implement the functionalities in current logic. 

There are also 2 critical changes that need to be made (not in this PR).
- avoid copying of spatialdata. This is really important, the object must not be copied, but only views should be passed in the function call.
- remove all transformations in a `pp` functionality. The way it's handled currently is quite confusing and happens in the `show` method.

giovp:
it's ready! it should have almost all functionalities as before + handle of caterogical and continuous for points

sagar87 (approved):

sagar87 (approved):

PR #50, opened 2023-04-12
Make _get_extent a utils function

timtreis:


src/spatialdata_plot/pl/basic.py
```diff
@@ -589,57 +557,79 @@ def show(
...
                     sdata.labels[key] = transform(sdata.labels[key], label_transformation)
 
-        # get biggest image after transformations to set ax size
-        x_dims = []
-        y_dims = []
+        # Get extent of data to be plotted
+        want_images_extent: bool = False
+        want_labels_extent: bool = False
+        want_points_extent: bool = False
+        want_shapes_extent: bool = False
```
giovp, line 564: think this should be a mapping, namedtuple, or dataclass

src/spatialdata_plot/pp/utils.py
```diff
@@ -47,14 +42,16 @@ def _verify_plotting_tree(sdata: sd.SpatialData) -> sd.SpatialData:
...
-    has_images = hasattr(sdata, "images")
-    has_labels = hasattr(sdata, "labels")
-    has_shapes = hasattr(sdata, "shapes")
+    has_images = True if sdata.images is not None else False
+    has_labels = True if sdata.labels is not None else False
+    has_shapes = True if sdata.shapes is not None else False
+    has_points = True if sdata.points is not None else False
 
     coordsys_keys = sdata.coordinate_systems
     image_keys = sdata.images.keys() if has_images else []
```
giovp, line 51:
```suggestion
    image_keys = [] if sdata.images is None else sdata.images.keys()
```

giovp: less verbose, can delete 45-48

src/spatialdata_plot/pl/utils.py
```diff
@@ -475,12 +625,19 @@ def _map_color_seg(
...
     unique_cols = np.unique(cols)
     if len(unique_cols) == 1 and unique_cols == 0:
         RNG = default_rng(42)
         cols = RNG.random((len(cols), 3))
 
+    # add alpha value to all colors ([0.1, 0.2, 0.3] -> [0.1, 0.2, 0.3, alpha])
+    target_alpha = 0.5
+    arr = np.array(cols)
+    new_col = np.full((arr.shape[0], 1), target_alpha)
+    cols = np.concatenate((arr, new_col), axis=1)
```
giovp, line 639: why is this needed?

src/spatialdata_plot/pl/basic.py
timtreis: Mapping would add more lines since it's a one-off thing, tuple would be immutable and dataclass also a little overkill for a one-off thing. Refactored to this instead

image

src/spatialdata_plot/pp/utils.py
timtreis: True, thanks!

src/spatialdata_plot/pl/utils.py
timtreis: Hoped at that point that it could convince the label2rgb function to produce transparent labels, turns out it doesn't. Removed it now 👍

PR #118, opened 2023-06-29
made render_images more robust to channel/cmap choices

timtreis:


src/spatialdata_plot/pl/basic.py
```diff
@@ -319,6 +321,8 @@ def render_images(
             Color to be used for NAs values, if present.
         alpha
             Alpha value for the shapes.
+        quantile_norm
```
LucaMarconato, line 324: which quantile?

LucaMarconato: Found in the private function, I would maybe expose the quantiles to the user via this function

LucaMarconato (approved):

src/spatialdata_plot/pl/basic.py
timtreis: It's based on https://github.com/CSBDeep/CSBDeep/blob/deb54947a734b959681ead6bde24c6143805663f/csbdeep/utils/utils.py#L51, don't know their exact reason behind it though. I could just link this function?

timtreis: > Found in the private function, I would maybe expose the quantiles to the user via this function

Good idea!

LucaMarconato: I would like the function yes.

timtreis:
Yeah, overhauled the entire `render_images` function because there were a bunch of edge cases for the permutations of the `channel`, `palette` and `cmap` options. @MobiTobi found a few bugs

timtreis:
I'll add tests to lock in this behaviour later and will then merge 👌

LucaMarconato:
Perfect thanks. One last question, how do you use images with tests? Do you assert that the plot needs to look like the images you put in the repo? If so, great idea, I need to do it as well when I transform data!

timtreis:
Yeah exactly, it's taken from Squidpy where @giovp uses that a lot. Super nice for end-to-end tests in plotting.

MobiTobi:
Looks good to me.

PR #119, opened 2023-06-29
Bugfix for `outline` parameter in `render_shapes`

timtreis:


src/spatialdata_plot/pl/utils.py
```diff
@@ -500,7 +500,13 @@ def _set_outline(
     gap_size = (point + (point * gap_width) * 2) ** 2
     bg_size = (np.sqrt(gap_size) + (point * bg_width) * 2) ** 2
     # the default black and white colors can be changes using the contour_config parameter
-    bg_color, gap_color = outline_color
+    if isinstance(outline_color, str):
```
LucaMarconato, line 503: the typing of `outline_color` in the signature should be adjusted to `tuple[str, str] | str | list[float]`. I would support also `tuple[list[float], list[float]]` for consistency.

LucaMarconato:
Nice! After addressing the comment on the argument type for me can be merged!

LucaMarconato (approved):

PR #124, opened 2023-06-30
"Color by column" now works again

timtreis:


LucaMarconato:
I just checked, in the spatial query notebook I get this:

image

While this was what we got in the past:
image

Is this expected?

timtreis:
Do you mean that both now use blue? I'm rather confused why the old one had two colors

LucaMarconato:
Yes. You are right, the all blue plot is actually the correct one.

LucaMarconato (approved):

PR #130, opened 2023-08-01
Can now plot columns from `GeoDataFrame` [BROKEN]

timtreis:


LucaMarconato:
I'd suggest adding a test for this for labels and one for labels, then I think it will be good to merge.

PR #152, opened 2023-08-30
`render_shapes` now has a `scale` parameter

timtreis:


timtreis:
@giovp This now scales the shapes. It however, doesn't adjust the coordinate axis extent. That'd be quite a lot of work since AFAIK we cannot easily use the transformation syntax for this since we need to scale each shape around its centroid and the transformation `scale` scales the layer.

Sonja-Stockhaus (approved):

LucaMarconato:
> This now scales the shapes. It however, doesn't adjust the coordinate axis extent. That'd be quite a lot of work since AFAIK we cannot easily use the transformation syntax for this since we need to scale each shape around its centroid and the transformation scale scales the layer.

I'd approximate it by identifying the largest radius of the collection of the shapes, let's say `r`, and, by calling `k` the scale factor, I'd enlarge the extent in each direction by `r*k - r`.

Notes:
- if all the shapes have the same radius then the extent is correct
- the approximation above is good enough, to be more precise one could use it only for large datasets, and use the exact (but more expensive) calculation for smaller dataset, described next.
- (obvious, but gonna be explicit for completeness) the exact calculation can be obtained by temporarily moving the radius column somewhere else, by scaling the radius column, by calling `get_extent()` and by replacing the radius column with the old one.
- (obvious, but gonna be explicit for completeness) the adjustment of the extent needs to be done only if the shapes are circles

PR #164, opened 2023-09-27
MultiscaleImage handling

Sonja-Stockhaus:
The user can either select a scale or the multi-scale images to be rendered are rasterized before

@timtreis, 2023-10-16: "Please also provide the first (pre-datashader) version of this notebook: https://github.com/scverse/spatialdata-notebooks/issues/59 "


timtreis (changes requested):
- Cover usecase in which I plot a list of `elements` and pass a list to `scale`, elementwise matching
- Implement logic which decouples img-raster DPI from canvas size DPI as discussed

LucaMarconato:
Just a quick comment, I have been using this new branch, works super smooth! Great work!

src/spatialdata_plot/pl/utils.py
```diff
@@ -1031,3 +1051,186 @@ def _mpl_ax_contains_elements(ax: Axes) -> bool:
...
+        target_unit_to_pixels = min(target_y_dims / y_dims, target_x_dims / x_dims)
+        image = rasterize(
+            image,
+            ("y", "x"),
+            [extent["y"][0], extent["x"][0]],
+            [extent["y"][1], extent["x"][1]],
+            coordinate_system,
+            target_unit_to_pixels=target_unit_to_pixels,
+        )
+        logging.info(
```
LucaMarconato, line 1153: I would remove this, I think it's enough for the user to experience fast performance and then highlight the performance improvement in a notebook.

src/spatialdata_plot/pl/utils.py
```diff
@@ -1031,3 +1051,186 @@ def _mpl_ax_contains_elements(ax: Axes) -> bool:
...
+            optimal_index_x -= 1
+
+        # pick the scale with higher resolution (worst case: downscaled afterwards)
+        optimal_scale = scales[min(optimal_index_x, optimal_index_y)]
+
+    # NOTE: problematic if there are cases with > 1 data variable
+    data_var_keys = list(multiscale_image[optimal_scale].data_vars)
+    image = multiscale_image[optimal_scale][data_var_keys[0]]
+
+    logging.info(f"Using scale '{optimal_scale}' from multi-scale image '{element}'.")
```
LucaMarconato, line 1234: Same as above, I would remove this. The user will get the plot fast, this message can't be used by the user in a actionable way. I would still highlight this in the documentation notebook.

Sonja-Stockhaus: Yeah, I get your point. Was probably most useful for me to see what was going on during coding and to demonstrate it in the documentation nb. It might help the user to see which scale is selected per default so they could better judge which scales might be too large. But would probably be rarely used

PR #188, opened 2023-11-29
Fix passing Colormap objects

aeisenbarth:
When passing a Colormap instance via `cmap`, an error is raised although it is supported according to documentation.

```
  File "…/site-packages/spatialdata_plot/pl/basic.py", line 370, in render_images
    cmap_params = _prepare_cmap_norm(
  File "…/site-packages/spatialdata_plot/pl/utils.py", line 339, in _prepare_cmap_norm
    cmap = copy(matplotlib.colormaps[rcParams["image.cmap"] if cmap is None else cmap])
  File "…/site-packages/matplotlib/cm.py", line 80, in __getitem__
    return self._cmaps[item].copy()
TypeError: unhashable type: 'ListedColormap'
```

This change looks up a colormap by name only if a string is passed.


timtreis:
Hey @aeisenbarth, (again :D) many thanks for your suggestion and the code contribution! @Sonja-Stockhaus has moved your changes over to https://github.com/scverse/spatialdata-plot/pull/194 so that we could add tests and slight modifications. So we'll close this one here now.

PR #190, opened 2023-11-29
Avoid dropping rendering commands

aeisenbarth:
When issueing multiple render commands of the same type, only the last command is executed. Multiple same rendering commands can be desired when blending different images (with alpha != 1) or creating a collage of images at different spatial locations.

This is caused by storing parameters in an ordered dictionary with the command type as key. Instead, onecould directly use a sequence of tuples.

### DIFF
--- a/src/spatialdata_plot/pl/basic.py
+++ b/src/spatialdata_plot/pl/basic.py
@@ -567,7 +567,7 @@ class PlotAccessor:
         ]
 
         # prepare rendering params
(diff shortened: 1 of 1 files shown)


timtreis:
Hey @aeisenbarth, many thanks for your suggestion and the code contribution! I have copied it over to https://github.com/scverse/spatialdata-plot/pull/192 so that I can easily add the corresponding tests. So I'll be closing this here.

PR #199, opened 2023-12-08
Updated render_shapes doc

timtreis:


src/spatialdata_plot/pl/basic.py
```diff
@@ -311,53 +505,121 @@ def render_points(
...
+                raise TypeError("Each item in 'quantiles_for_norm' must be a float or int within [0, 100], or None.")
+
+            pmin, pmax = quantiles_for_norm
+            if pmin is not None and pmax is not None and pmin > pmax:
+                raise ValueError("The first number in 'quantiles_for_norm' must not be smaller than the second.")
+
+        if scale is not None and not isinstance(scale, (list, str)):
+            raise TypeError("Parameter 'scale' must be a string or a list of strings.")
+
+        if isinstance(scale, list) and not all(isinstance(s, str) for s in scale):
```
Sonja-Stockhaus, line 620: will this break in case s is None? Bc according to the doc we allow it

src/spatialdata_plot/pl/basic.py
```diff
@@ -398,52 +660,55 @@ def render_images(
...
+        norm : Normalize | None, optional
             Colormap normalization for continuous annotations, see :class:`matplotlib.colors.Normalize`.
-        na_color
+        na_color : ColorLike | None, optional
             Color to be used for NAs values, if present.
-        alpha
+        outline_alpha : float | int, default 1.0
+            Alpha value for the labels.
+        fill_alpha : float | int, default 0.3
             Alpha value for the labels.
```
Sonja-Stockhaus, line 710: maybe add outline or fill respectively so that we don't have the same description twice?

src/spatialdata_plot/pl/basic.py
```diff
@@ -459,6 +724,66 @@ def render_labels(
...
+            raise TypeError("Parameter 'outline_alpha' must be numeric.")
+
+        if not isinstance(fill_alpha, (float, int)):
+            raise TypeError("Parameter 'fill_alpha' must be numeric.")
+
+        if scale is not None:
+            if not isinstance(scale, (list, str)):
+                raise TypeError("Parameter 'scale' must be a string or a list of strings.")
+            scale = [scale] if isinstance(scale, str) else scale
+            if not all(isinstance(s, str) for s in scale):
```
Sonja-Stockhaus, line 784: same as above, None also allowed?

Sonja-Stockhaus (approved):
Really nice!!

src/spatialdata_plot/pl/basic.py
timtreis: No, doesn't break. `None` toggles neither L617 or L620, so it's evaluated correctly later on 👍

src/spatialdata_plot/pl/basic.py
timtreis: Yeah, `None` is allowed. I didn't include it in the error message because it only ever shows when the user meddles with it anyway 🤔

PR #201, opened 2023-12-12
Plotting of continuous values when norm=None

Sonja-Stockhaus:
Add a case to set norm when `norm=None` in `utils._prepare_cmap_norm`


Sonja-Stockhaus:
@timtreis could you please have a look at the bugfix? It works now, but there was still an open TODO in the code and I'm not sure what the intention for the original first case was...

LucaMarconato:
Amazing @Sonja-Stockhaus! Thanks for the fix 😊 Since the todo was there also before the fix I would merge now and eventually circle back later.

Sonja-Stockhaus:
@LucaMarconato okay, thanks for the feedback!

timtreis (approved):

PR #208, opened 2024-01-11
Fix "xarray has no attribute reshape"

aeisenbarth:
A `.values` attribute has been forgotten.

`label` is a SpatialImage and DataArray. Scikit-image expects a true array with a reshape method. All other uses of `_map_color_seg` pass the correct value, e.g.:
https://github.com/scverse/spatialdata-plot/blob/09a31adec012483d1bc578a1fa4802ce8e8b8046/src/spatialdata_plot/pl/render.py#L618-L620

### DIFF
--- a/src/spatialdata_plot/pl/render.py
+++ b/src/spatialdata_plot/pl/render.py
@@ -663,7 +663,7 @@ def _render_labels(
         else:
             # Default: no alpha, contour = infill
             label = _map_color_seg(
-                seg=label,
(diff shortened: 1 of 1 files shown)


aeisenbarth:
(I thought I had seen this already fixed by another contributor, but couldn't find it merged or in a PR.)

timtreis:
True, thank you!

timtreis (approved):

PR #215, opened 2024-02-14
Categorical Legend Fix

### LINKED ISSUE #214: Categorical Data is incorrectly colored in the legend when plotting
When using `pl.render_labels` with the `color` and `palette` parameter, the legends are incorrectly colored. 

Example Plot:
NBL-25-R2C8--Diagnosis--incorrect

Here we see that the Dot plot from `scanpy` counts 1.1K cells in the cluster `NBL_9` and colors them with pastel green (correct color). However on the `spatialdata` plots, cluster `NBL_11` is pastel green, when it should be `NBL_9`. The colors on the `spatialdata` image itself are not wrong, it is the colors on the legend which are incorrect.

This issue is occurs in `pl.utils._get_palette`, where the values in the`categories` gets mapped to the colors of the palette. However in `scanpy` it's the categories of the DataFrame which get mapped to the colors.

By adjusting the following line from https://github.com/scverse/spatialdata-plot/blob/595ea44e2a7414e23d9967ea180d54f4af467b2d/src/spatialdata_plot/pl/utils.py#L753-L754

to `dict(zip(categories.categories, palette))` we get the correct legend-color mapping.

NBL-25-R2C8--Diagnosis--correct

## Reproducible Example - MIBITOF dataset

```python
import spatialdata as sd
import spatialdata_plot
from pathlib import Path

mibitof_sd = sd.read_zarr(Path("~/Downloads/data.zarr").expanduser())

fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(18, 8), layout="tight")

mibitof_sd.pp.get_elements("point8_labels").pl.render_labels(color="Cluster").pl.show(
    ax=axes[0]
)

mibitof_sd.pp.get_elements("point8_labels").pl.render_labels(
    color="Cluster", palette=["blue"], groups=["Epithelial"]
).pl.show(ax=axes[1])

fig
```

Generated Plot: Here `Imm_other` in the legend actually represents the `Epithelial` cells.
point8_labels

srivarra:
Closes #214.

When plotting the SpatialData labels element, the colors are incorrectly labeled in the legend of the figure when `palette` and `color` parameters are used in the `pl.render_labels` method.

This is due to `pl.utils._get_palette` using the values in the `categories` (`pd.Categorical`) DataFrame, instead of the *categories* of `categories`.

Here is point8 from the MIBITOF dataset with the corrected color-label legend mappings.
point8_labels

### DIFF
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -804,8 +804,8 @@ def _decorate_axs(
         # Adding legends
         if is_categorical_dtype(color_source_vector):
             # order of clusters should agree to palette order
-            clusters: pd.Categorical = color_source_vector.unique()
-            clusters: pd.Categorical = clusters[~clusters.isnull()]
+            clusters_unique: pd.Categorical = color_source_vector.unique()
+            clusters: pd.Categorical = clusters_unique[~clusters_unique.isnull()]
             palette = _get_palette(
                 adata=adata, cluster_key=value_to_plot, categories=clusters, palette=palette, alpha=alpha
             )
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -713,6 +713,7 @@ def _get_palette(
     palette: ListedColormap | str | list[str] | None = None,
     alpha: float = 1.0,
 ) -> Mapping[str, str] | None:
+    categories = categories.categories
     if adata is not None and palette is None:
         try:
             palette = adata.uns[f"{cluster_key}_colors"]  # type: ignore[arg-type]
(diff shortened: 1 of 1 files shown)


melonora:
Hello @srivarra,

Thanks for your contribution! We are pushing hard for a release that will allow for multiple tables in SpatialData, but we will get to your PR soon:)

melonora:
Hello again,

It turns out that with the work that I was doing for adopting this repo to the new multi_table implementation this got fixed along with that work and thus this PR is not required anymore.
image

image

Thank you though for pointing out the issue!

melonora:
Note that the carcinoma example as it is with the current state of the data all values are carcinoma. I double checked this because you seemed to have this differently.

PR #221, opened 2024-03-05
Fix channel str support

berombau:
These test cases did not work:
```python
    def test_plot_can_render_a_single_channel_str_from_image(self, sdata_blobs_str: SpatialData):
        sdata_blobs_str.pl.render_images(elements="blobs_image", channel="c1").pl.show()

    def test_plot_can_render_a_single_channel_str_from_multiscale_image(self, sdata_blobs_str: SpatialData):
        sdata_blobs_str.pl.render_images(elements="blobs_multiscale_image", channel="c1").pl.show()

    def test_plot_can_render_two_channels_str_from_image(self, sdata_blobs_str: SpatialData):
        sdata_blobs_str.pl.render_images(elements="blobs_image", channel=["c1", "c2"]).pl.show()

    def test_plot_can_render_two_channels_str_from_multiscale_image(self, sdata_blobs_str: SpatialData):
        sdata_blobs_str.pl.render_images(elements="blobs_multiscale_image", channel=["c1", "c2"]).pl.show()
```

I added the tests and the two fixes needed to make them pass. Reported by @thienthu.
The problem with buggy channel string support is usually creating a new Image2DModel with the data but not also the string labels via a c_coords parameter (see `src/spatialdata_plot/pl/utils.py:1247`).


timtreis:
Should now all be fixed, it's failing because the codecov retry limit is reached for now. Will try again in a bit and then merge. Thanks for the work @berombau!

PR #235, opened 2024-03-31
Mypy

melonora:
Pr to fix the horrors of mypy. Will be followed by a PR for docs pass.


LucaMarconato:
Omg 🙈 Internet is quite bad at the moment, I can use git from the command line but I can't review with the GitHub interface. I'll check this when I land or tomorrow. 

But from the commit messages it seems like a tremendous effort, well done 😁

melonora:
I will probably merge it already though, or I can just create a bunch of follow up prs. Regarding effort, most problematic thing is mypy being very very cryptic with its messages and the typing changing from one part to the next.

melonora:
we started with about160 mypy messages

PR #236, opened 2024-04-01
Fix var plotting

### LINKED ISSUE #234: Plotting var on shapes broken
Plotting from var on shapes spotted working, this is probably introduced with the latest refactoring and probably linked to the same reason behind the previously reported bug: https://github.com/scverse/spatialdata-plot/issues/228.

melonora:
closes #234 
Follow up on #235 fixing the plotting color columns that are numerical. First this fixes the visibility of the colorbar when plotting labels with a color. The issue here was that due to the outline_alpha being set to 0 when outline is False and outline of the label being plotted after the labels it would overwrite the colorbar. This was fixed by first plotting outline and then the label. A potential issue here is that when both alpha and outline alpha are 1 the outline will not be visible, but currently I thought this best as there is no clear way to decouple the colorbar from the outline. The general rule here is whatever is plotted last is visible.

The other issue got reported by @LucaMarconato  in #234 . Ultimately, this was an issue when plotting `feature0` in the `models2` notebook in `spatialdata-notebooks`. The issue here was that 3 shapes are being plotted, one multipolygon and 2 polygons. This results in a `color_vector` of length 3 which is convertible to a RGB color, causing 1 color to be created for 3 shapes. This PR fixes the issue by checking whether we are dealing with a 1 dimensional color vector of length 3 or 4 and whether it then has the same length as the number of shapes.

The plot in `models2` would now look like this:
afbeelding

### DIFF
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -218,7 +218,7 @@ def _get_collection_shape(
                         " categories, set the column to categorical dtype."
                     ) from e
                 c = cmap(norm(c))
-        elif not c:
+        else:
             fill_c = ColorConverter().to_rgba_array(c)
     except ValueError:
         if norm is None:
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -1970,7 +1970,7 @@ def _is_coercable_to_float(series: pd.Series) -> bool:
 def _return_list_str_none(parameter: list[str | None] | str | None) -> list[str | None]:
     """Force mypy to recognize list of string and None."""
     if isinstance(parameter, list) and all(isinstance(item, (str, type(None))) for item in parameter):
-        checked_parameter = parameter if isinstance(parameter, list) else [None]
+        checked_parameter = parameter
     else:
         checked_parameter = [None]
     return checked_parameter
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -839,7 +839,7 @@ def _decorate_axs(
     ax: Axes,
     cax: PatchCollection,
     fig_params: FigParams,
-    value_to_plot: str | None,  # str | None,
+    value_to_plot: str | None,
     color_source_vector: pd.Series[CategoricalDtype],
     adata: AnnData | None = None,
     palette: ListedColormap | str | list[str] | None = None,
@@ -1478,7 +1478,7 @@ def _validate_colors_element_table_mapping_points_shapes(
                         params.col_for_color.append(col_color)
                         element_table_mapping[element_name] = set()
                     else:
-                        if isinstance(mapping := element_table_mapping[element_name], set) and len(mapping.copy()) != 0:
+                        if isinstance(mapping := element_table_mapping[element_name], set) and len(mapping) != 0:
                             for table_name in mapping.copy():
                                 if (
                                     col_color not in sdata[table_name].obs.columns
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -206,7 +206,20 @@ def _get_collection_shape(
 
     try:
         # fails when numeric
-        fill_c = ColorConverter().to_rgba_array(c)
+        if len(c.shape) == 1 and c.shape[0] in [3, 4] and c.shape[0] == len(shapes) and c.dtype == float:
+            if norm is None:
+                c = cmap(c)
+            else:
+                try:
+                    norm = colors.Normalize(vmin=min(c), vmax=max(c))
+                except ValueError as e:
+                    raise ValueError(
+                        "Could not convert values in the `color` column to float, if `color` column represents"
+                        " categories, set the column to categorical dtype."
+                    ) from e
+                c = cmap(norm(c))
+        elif not c:
+            fill_c = ColorConverter().to_rgba_array(c)
     except ValueError:
         if norm is None:
             c = cmap(c)
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -1478,13 +1478,13 @@ def _validate_colors_element_table_mapping_points_shapes(
                         params.col_for_color.append(col_color)
                         element_table_mapping[element_name] = set()
                     else:
-                        if isinstance(mapping := element_table_mapping[element_name], set) and len(mapping) != 0:
-                            for table_name in mapping.copy():
+                        if isinstance(table_set := element_table_mapping[element_name], set) and len(table_set) != 0:
+                            for table_name in table_set.copy():
                                 if (
                                     col_color not in sdata[table_name].obs.columns
                                     and col_color not in sdata[table_name].var_names
                                 ):
-                                    mapping.remove(table_name)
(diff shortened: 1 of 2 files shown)


melonora:
Test that fails on 3.10 is due to images not matching. However, on 3.9 it matches so not certain if we can really fix that. For the record this is the output:
afbeelding

melonora:
lovely, tests in #237 do pass so we can ignore the failing test here.

giovp (approved):
ok for me ~~but maybe a very small test that would have caught this?~~ ah nvm ok with #237

LucaMarconato:
Thanks @melonora for the PR! I noticed that neither main neither this PR is able to run the `technology_mibitof.ipynb` notebook because of a plot of a categorical into labels that fails.

I see a similar test is failing in this PR, so probably worth fixing them both, could you have a look at it please? After this we can merge.

melonora:
hmm wait this PR was fine as the follow up PR the tests just passed. However with the last change on the follow up PR it is failing

src/spatialdata_plot/pl/utils.py
```diff
@@ -822,7 +839,7 @@ def _decorate_axs(
     ax: Axes,
     cax: PatchCollection,
     fig_params: FigParams,
-    value_to_plot: str | None,
+    value_to_plot: str | None,  # str | None,
```
timtreis, line 842: remove

src/spatialdata_plot/pl/utils.py
```diff
@@ -1379,46 +1398,56 @@ def _create_initial_element_table_mapping(
                     warnings.warn(
                         f"The element '{element}' is not annotated by table '{table_name}'", UserWarning, stacklevel=2
                     )
-                element_table_mapping[element].add(table_name)
+                if isinstance(mapping := element_table_mapping[element], set):
+                    mapping.add(table_name)
+    assert isinstance(element_table_mapping, dict)
```
timtreis, line 1403: I personally dislike spurious asserts. Can you verify before and block this path preemptively? If not, rather throw an informative warning

src/spatialdata_plot/pl/utils.py
```diff
@@ -1379,46 +1398,56 @@ def _create_initial_element_table_mapping(
...
 
 def _update_element_table_mapping_label_colors(
     sdata: SpatialData, params: LabelsRenderParams | PointsRenderParams | ShapesRenderParams, render_elements: list[str]
 ) -> ImageRenderParams | LabelsRenderParams | PointsRenderParams | ShapesRenderParams:
-    element_table_mapping = params.element_table_mapping
+    element_table_mapping: dict[str, set[str | None] | str | None] | str | list[str] | None = (
+        params.element_table_mapping
+    )
+
+    assert isinstance(element_table_mapping, dict)
```
timtreis, line 1415: same

src/spatialdata_plot/pl/utils.py
```diff
@@ -1427,8 +1456,13 @@ def _update_element_table_mapping_label_colors(
 def _validate_colors_element_table_mapping_points_shapes(
     sdata: SpatialData, params: PointsRenderParams | ShapesRenderParams, render_elements: list[str]
 ) -> PointsRenderParams | ShapesRenderParams:
-    element_table_mapping = cast(dict, params.element_table_mapping)
-    if len(params.color) == 1:
+    element_table_mapping: dict[str, set[str | None] | str | None] | str | list[str] | None = (
+        params.element_table_mapping
+    )
+
+    assert isinstance(element_table_mapping, dict)
```
timtreis, line 1463: same

src/spatialdata_plot/pl/utils.py
```diff
@@ -1444,13 +1478,13 @@ def _validate_colors_element_table_mapping_points_shapes(
                         params.col_for_color.append(col_color)
                         element_table_mapping[element_name] = set()
                     else:
-                        if len(element_table_mapping[element_name].copy()) != 0:
-                            for table_name in element_table_mapping[element_name].copy():
+                        if isinstance(mapping := element_table_mapping[element_name], set) and len(mapping.copy()) != 0:
```
timtreis, line 1481: reads a bit code-golf-y 👀 Also, why are you copying? Theoretically, you could also remove the `!= 0` since the len of an empty set is also 0 (False)

src/spatialdata_plot/pl/utils.py
```diff
@@ -1444,13 +1478,13 @@ def _validate_colors_element_table_mapping_points_shapes(
...
+                            for table_name in mapping.copy():
```
timtreis, line 1482: These are shallow copies no? Why do you do that?

src/spatialdata_plot/pl/utils.py
```diff
@@ -1470,24 +1504,28 @@ def _validate_colors_element_table_mapping_points_shapes(
             params.color = [None] * len(render_elements)
             params.col_for_color = [None] * len(render_elements)
         else:
+            assert isinstance(params.color, list)
+            assert isinstance(params.col_for_color, list)
```
timtreis, line 1508: same

src/spatialdata_plot/pl/utils.py
```diff
@@ -1886,20 +1945,47 @@ def _get_wanted_render_elements(
...
-def _is_coercable_to_float(series):
+def _is_coercable_to_float(series: pd.Series) -> bool:
     numeric_series = pd.to_numeric(series, errors="coerce")
     return not numeric_series.isnull().any()
+
+
+def _return_list_str_none(parameter: list[str | None] | str | None) -> list[str | None]:
+    """Force mypy to recognize list of string and None."""
+    if isinstance(parameter, list) and all(isinstance(item, (str, type(None))) for item in parameter):
+        checked_parameter = parameter if isinstance(parameter, list) else [None]
```
timtreis, line 1973: You already checked whether it's a list?

timtreis (changes requested):
very few minor things

src/spatialdata_plot/pl/utils.py
melonora: This is purely due to mypy not accepting anything else. The only way to get around this is by adopting the whole code base to use params per element.

src/spatialdata_plot/pl/utils.py
melonora: mypy...

src/spatialdata_plot/pl/utils.py
melonora: Mypy gets very confused because of the data structure constantly changing. I tried casting but in most cases this does not help

src/spatialdata_plot/pl/utils.py
melonora: again partially mypy as for the copy, if you remove from a set during a for loop you will get an error because indices are removed while looping over the object. Here the one in `len` is unnecessary though so removing that.

src/spatialdata_plot/pl/utils.py
melonora: you can try remove it and see what mypy does:P

melonora: ah yeah initially I had an issue with mypy not accepting it but that could have been due to another conflict. Now it passes when removing.

melonora:
Merging after talking with @timtreis

PR #237, opened 2024-04-01
Pass docs

melonora:
Follow up on #236. This PR is a pass on the docs. The type hints have been updated to the last status. Currently pl.show is missing some docstrings, but due the behaviour not being completely correct I kept these arguments out for now.


src/spatialdata_plot/pl/basic.py
```diff
@@ -291,17 +302,28 @@ def render_points(
         elements : list[str] | str | None, optional
             The name(s) of the points element(s) to render. If `None`, all points
             elements in the `SpatialData` object will be used.
-        color : Colorlike | str | None, optional
-            Can either be a color-like or a key in :attr:`sdata.table.obs`. The latter
-            can be used to color by categorical or continuous variables.
+        color : list[str | None] | str | None
+            Can either be string(s) representing a color-like or key(s) in :attr:`sdata.table.obs`. The latter
```
giovp, line 306: I don't think it can be stuff in adata.obs, as points are not annotated by tables

giovp: also below, is this correct? I don't think so,I think it was copy pasted?

giovp (approved):
only a minor thing on docstrings, but it would be nice to also have real world example (small ones, self contained, maybe with blobs datasets)to explain the different logic cause it gets quite complicated (list of lists?)

src/spatialdata_plot/pl/basic.py
melonora: hehe points now can be annotated by tables and I agree on tests, still have to write them. In general also a lot of the notebooks have plots that have no equivalent tests here so before a lot of them were not properly showing either. However, for release it was pushed. The logic of this library is quite complicated with some parameters with the same name are used in different ways that it took me quite some time so at the moment I did the testing by checking whether all the notebook plots now display correctly.

LucaMarconato:
Thanks, merging now. Please note that specifying the type and the default values after the parameters list in the docstring is not required since this is automatically inferred by sphinx. Not worth to remove the old ones, but I'd suggest not to keep adding them in new functions. @melonora @timtreis. CC @giovp

PR #243, opened 2024-04-03
basic shapes datashader rendering

ckmah:
Provides basic datashader support for ShapeElements. Still needs to address other parameters (outlines an such), but proof of concept works!

Example rendering 2389 polygons from merfish dataset:

Using matplotlib (22.4s):
image

Using datashader (6.6s):
image


LucaMarconato:
Woah! 🚀 The results and performance look great! Merging this to @Sonja-Stockhaus's branch and then we can do a joint review there.

PR #245, opened 2024-04-05
_set_color_source_vec and _render_points Refactor 

dylanclam12:
Refactored _set_color_source_vec to be compatible with the new get_values that can grab table annotating points.
Also refactored _render_points to be able to plot the new annotations.


timtreis:
@dylanclam12 I think the codebase has changed quite a bit since you opened this draft - shall I close ?

LucaMarconato:
@timtreis since the code is short I would still try to merge or see if all the functionalities are ported. Before `get_values()` was not used for points leading to code duplication.

PR #258, opened 2024-05-16
Fix tests in main

timtreis:


timtreis:
Knowingly merged incorrect images, but wanted to get back to tests on main passing. Documented incorrect images here: https://github.com/scverse/spatialdata-plot/issues/259

PR #262, opened 2024-05-17
Categorical data is now correctly colored in the legend when plotting

timtreis:


LucaMarconato:
The branch can be deleted right? @timtreis

PR #267, opened 2024-05-28
Image render param refactor

melonora:
This PR refactors the way the parameters for image rendering get validated and passed. Some of the changes are not backward compatible. For `element` the user can't pass on a `list` of `str` anymore, but only `str` or `None`. All the passed on parameters are used to create a view config for a single element. If `element` is `None` then all the parameters are checked for whether they are valid for a specific element and if that is not the case the value of the parameter for that element is set to None. If a user wants to plot multiple specific images, the `pl.render` calls can be chained. 

Because of creating viewconfigs in such a way that each element has its own contained config this will make it easy to hook up vega like specifications and also to write those out.

### DIFF
--- a/src/spatialdata_plot/_utils.py
+++ b/src/spatialdata_plot/_utils.py
@@ -1,0 +1,69 @@
+from __future__ import annotations
+
+import functools
+import warnings
+from typing import Any, Callable, TypeVar
+
+RT = TypeVar("RT")
+
+
+def deprecation_alias(**aliases: str) -> Callable[[Callable[..., RT]], Callable[..., RT]]:
+    """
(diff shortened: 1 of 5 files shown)


melonora:
These tests pass locally

melonora:
ok couple of things that "fail":
This PR versus reference:
afbeelding
afbeelding
Again this PR versus reference:
afbeelding
afbeelding

afbeelding
afbeelding

The remainder of the tests were thinking at least locally for me also on main. @timtreis what is the current status of #262? I would have to merge that one in here, but right now I have no permission to actually checkout that PR with the github cli.

timtreis (approved):
Only known-to-fail tests fail. We'll fix in main after this refactor

PR #275, opened 2024-06-12
Fix transformed labels with outline

### LINKED ISSUE #273: Label outlines cause wrong transformation
It seems `outline=True` takes a code path that causes labels to have a wrong scale.

It is important for the images that I render to not have filled labels, but outlines.

## Example

```python
import numpy as np
import spatialdata as sd
import spatialdata_plot # noqa: F401
from skimage.transform import AffineTransform

# Same array, same transformation used for both image and labels.
# They should exactly overlap.
array = np.zeros((100, 100), dtype=int)
array[30:70, 30:70] = 1
affine = sd.transformations.Affine(
    AffineTransform(translation=(10, 10), rotation=np.deg2rad(15), scale=0.65).params,
    input_axes=("y", "x"),
    output_axes=("y", "x"),
)
image1 = sd.models.Image2DModel.parse(
    array[:, :, np.newaxis], dims=("y", "x", "c"), transformations={"global": affine}
)
labels1 = sd.models.Labels2DModel.parse(
    array, dims=("y", "x"), transformations={"global": affine}
)
sdata = sd.SpatialData(images={"image1": image1}, labels={"labels1": labels1})

# Render with defaults. Spatialdata-plot fills the labels with color, but no outline.
sdata.pl.render_images("image1").pl.render_labels("labels1").pl.show(title="without outline", save="without_outline.png")

# Render without fill, but with outline, because I don't want to occlude what is behind the labels.
sdata.pl.render_images("image1").pl.render_labels(
    "labels1", fill_alpha=0.0, outline=True, outline_alpha=1.0
).pl.show(title="with outline", save="with_outline.png")
```

aeisenbarth:
Outlines are rendered on an axis separate from the filled labels. The transformation was only applied to the filled labels, so that the outlines had an identity transform. This was not observable when outlines were off or the transform was an identity (as in the blobs data).

Closes #273

For comparison with the example in the linked issue:


aeisenbarth:
I looked into the 2 lines of missing coverage. These are the two lines that seemed to be wrongly indented. By indenting them, I moved them into an untested code block, so that they are not covered anymore.

The else-branch is commented as "Default: no alpha, contour = infill" and not covered by tests. I tried adding a test, but am not convinced that this is the correct result. I think the labels should be filled and the test should fail. That means the test can only be added if someone has a plot file of the correct, expected result. Or the else-branch should be removed.
image

<details>
<summary>Test case for infill</summary>

```python
    def test_plot_can_render_contour_with_infill(self, sdata_blobs):
        # Rendering with fill_alpha == outline_alpha takes a special code path
        sdata_blobs.pl.render_labels(
            "blobs_labels",
            color="channel_0_sum",
            fill_alpha=1.0,
            outline=True,
            outline_alpha=1.0,
        ).pl.show()
```

</details>

melonora:
Hi @aeisenbarth, thanks for the PR. I will merge a couple of PRs first and then adjust this one if required.

timtreis:
@aeisenbarth can you rename that one function?

image

PR #280, opened 2024-06-22
implement label fix

melonora:
Labels were still being plotted even when `wanted_elements_on_this_cs` was an empty list due to this particular code block not being in the if conditional block. This test fixes that.


melonora:
bypassing for now as we require a spatialdata release for tests to pass.

PR #295, opened 2024-07-13
Added benchmarks for plots

LucaMarconato:
I added some basic sets of benchmarks for measuring the time to produce plots. The correctness of the plots is currently not checked, we can modify it later.

The only thing missing from the PR is the integration with GitHub Actions, so that we will catch on time eventual future performance drops. To finalize that I need some permissions from scverse.


melonora:
you mean for the token?

LucaMarconato:
Good news I setup everything. Bad news, CodSpeed doesn't seem to be suitable for the benchmarking use cases of `spatialdata-plot` because [it excludes system calls by design](https://docs.codspeed.io/features/understanding-the-metrics/#system-calls), and we need them when looking at plotting performance. 

See for example this screenshot; the execution time for `test_plot_can_render_large_image[image]` is 54.2s, but CodSpeed reports 273.7 milliseconds because most of the time is spent on system calls:
image

I wrote to the devs to see if there is an option to get around this, but most likely I'll end up closing this PR and switching to `asv`; which btw @Czaki is setting up on `napari-spatialdata` https://github.com/scverse/napari-spatialdata/pull/289 right today, good timing!

LucaMarconato:
I archived the branch behind this PR here https://github.com/scverse/spatialdata-plot/releases/tag/archive%2Fcodspeed_benchmarks, I'll delete this branch.

Issue #242, opened 2024-04-03
modify colormap so that 0 values have alpha=0


giovp:
saw this from @LucaMarconato tutorial on visium HD https://github.com/scverse/spatialdata-notebooks/pull/93

```python
import matplotlib.cm as cm
import matplotlib.pyplot as plt
import numpy as np
from matplotlib.colors import LinearSegmentedColormap

# modify the viridis colormap, so that the top color is a green (better visible on the H&E pink), and such that
# the value 0 leads to a transparent color
viridis = cm.get_cmap("viridis", 256)
# using 0.8 instead of 1.0 truncates the colormap
colors = viridis(np.linspace(0, 0.8, 256))
# set the color of zero to be transparent
colors[0, :] = [1.0, 1.0, 1.0, 0.0]

new_cmap = LinearSegmentedColormap.from_list("truncated_viridis", colors)
```

it's actually really nice and result to this

![image](https://github.com/scverse/spatialdata-plot/assets/25887487/1ce0a843-fed0-40b2-9e32-5d05771d9d2c)

I think it could just be an option to do this modification internally? with an argument like `alpha_value_zero: bool = True` or so, wdyt?

grst:
Could also be a helper function that takes an arbitrary `cmap` and returns a modified one? Personally, I'm not a big fan of the plotting functions accumulating too many arguments. 

I've been using something similar for visium data actually... this version makes a gradient from transparent to viridis: 

```python
def cmap_alpha(cmap="viridis_r"):
    cmap = cm.get_cmap("viridis_r")
    colors = cmap(np.arange(cmap.N))
    colors[: int(cmap.N / 2), 3] = np.linspace(0, 1, int(cmap.N / 2))  # Modify alpha channel
    colors[int(cmap.N / 2) :, 3] = 1  # Modify alpha channel
    cmap = ListedColormap(colors)
    return cmap
```

timtreis:
Wouldn't this be a case in which `na_color` should do the trick, at least semantically?

timtreis:
I do see the usecase but I'm not a fan of putting 0 as "missing", in this case I think it'd be cleaner to set those to `NA` and hide them through `na_color`. A hacky way would be to instead expose an option that modifies a given cmap so that the gradient towards 0 is not just towards white but transparent white?

Bioin-Mixologist:
hello, @giovp @timtreis . I used the same code according to the tutorial, but the result is not ideal, is it the version reason?
[...]

PR #302, opened 2024-07-14
Added utils function for 0-transparent cmaps

timtreis:
Note: the only existing utils function was using a different testing pattern, which I now aligned with the usual one. But therefore there's an additonal image in this PR


LucaMarconato:
Amazing! Ready for review?

timtreis:
Added the function @LucaMarconato @giovp. Probably best to add it to one of the notebooks to illustrate usage - which one do you think would be best?

LucaMarconato:
Thanks! I am working on the Visium HD notebook right now, so I can add it there 😊

melonora (approved):
LGTM! Maybe one consideration is to add a parameter that automatically does this for the user e,g, `zero_transparent`, but this could be added in a different PR. Also this PR does make it a bit more difficult for view configurations as we are dealing with different alpha values. May have to think of serialization of the colormap values.

timtreis:
Yeah, based on https://github.com/scverse/spatialdata-plot/issues/242 we decided not to include it as an argument but like this. But I agree, eventually, we should probably just serialise the data. I think ggplot f.e. also does it like this when generating the plot

PR #307, opened 2024-07-15
Fix setting vmin vmax

### LINKED ISSUE #303: Unable to set vmin and vmax when plotting raster data
I am trying to plot some raster data with a different normalization (vmin, vmax) for the cmap. Both approaches I used unfortunately don't work, can you have a look at this please?

```python
import spatialdata_plot
import matplotlib.colors

sdata = blobs()

sdata['single_channel'] = sdata['blobs_image'].sel(c=slice(0, 0))

# original
sdata.pl.render_images('single_channel').pl.show()
plt.show()

# attempt 1, doesn't work
sdata.pl.render_images('single_channel', vmin=0, vmax=0.4).pl.show()
plt.show()

# attempt 2, doesn't work
norm = matplotlib.colors.Normalize(vmin=0, vmax=0.4)
sdata.pl.render_images('single_channel', norm=norm).pl.show()
plt.show()
```

melonora:
closes #303 

Fixes the problem in #303. Main problem is that clipping for the `Normalize` object is set to `False` by default in `matplotlib`. I have set this value to `True` and added tests. I don't think we currently have a usecase for setting it to `False`, but feel free to comment if there is a usecase for us currently.


timtreis (changes requested):
Code lgtm, but can you add the fix to the changelog? Also, what do you think of plotting a 2x1 grid with the non-clipped and clipped image so it's visually easy see the difference?

melonora:
> Code lgtm, but can you add the fix to the changelog? Also, what do you think of plotting a 2x1 grid with the non-clipped and clipped image so it's visually easy see the difference?

added to changelog. Regarding 2 x 1 grid, I would if ncols was working xD. I don't want to create a fig object here now.

timtreis:
Why the aversion against the fig object? Used it f.e. in the 0-transparency cmap PR: https://github.com/scverse/spatialdata-plot/pull/302/files#diff-59872f651a8fef967621d774a8b7441219dc474f1d9604424c54fe19a162bbfd

melonora:
I would just solve the issue with ncols as well:) like I would not add more code in a test because of ncols, though behavior in this case is a bit tricky

melonora:
added the baseline to the test for now and will open an issue on ncols as argument

timtreis (approved):
LGTM, thanks for adjusting the test!

LucaMarconato (approved):

Issue #296, opened 2024-07-13
Worse performance using datashader?


LucaMarconato:
I wrote some benchmarks available here https://github.com/scverse/spatialdata-plot/pull/295 (they can simply run as tests) and I have noticed that the datashader performance is worse than the matplotlib based one.

I think this maybe be due to the size of the canvas used by datashader since in the MERFISH example here https://github.com/scverse/spatialdata-plot/pull/243 the performance was (as expected) better. 

Therefore using a smaller default canvas size may fixed the issue. @Sonja-Stockhaus could you please have a look into this?

LucaMarconato:
Here are the results of a (single) run of the tests (the timing are consistent across multiple manual runs).

<img width="551" alt="image" src="https://github.com/user-attachments/assets/3fd02496-686b-4910-a8bc-6ca470f2b49c">

LucaMarconato:
With the fix that I proposed to the performance bug here https://github.com/scverse/spatialdata-plot/issues/297 the performance gap is much bigger

<img width="548" alt="image" src="https://github.com/user-attachments/assets/ea5ae476-8b5d-4f48-8dea-d2028586dfc3">

timtreis:
@Sonja-Stockhaus my "didn't-look-at-the-code" theory is that datashader generates too large of an image which then bypasses the rasterisation-downsampling logic. Wdyt?

Sonja-Stockhaus:
Yep, datashader generates an image that is exactly the size of the extent (large extent = large image = long runtime). I'll think of sth so that we can use a smaller canvas size and then maybe rasterize or so to bring it back to the original scale.
Do we want a heuristic again to decide on the "smaller canvas size"?

I also noticed that for datashader, e.g. the radius of the points is relative to the axes which is not the case for matplotlib. So for a large extent you need extremely large point sizes to even make them visible at all with datashader. That should be consistent with matplotlib.

LucaMarconato:
Thanks for the explanation. I would reuse the logic of `_rasterize_if_necessary()` or `_multiscale_to_spatial_image()` to take the `dpi` of the figure and the `fig_size` into consideration, since the extent could be extremely large, but in the end we are limited by the pixels available on screen/paper for plotting.

LucaMarconato:
[...]

Issue #291, opened 2024-07-13
Error in Squidpy notebook due to rasterization not handling transformations


LucaMarconato:
The [Squidpy notebook](https://github.com/scverse/spatialdata-notebooks/blob/main/notebooks/examples/technology_visium_hd.ipynb) shows an error liked to rasterizatation. @Sonja-Stockhaus could you please have a look at it? Thanks 😊 

This is part of the log:
```
    300 else:
    301     array_module = np
--> 302 return lambda shape: tuple(c(shape, array_module) for c in creators)
File /mnt/miniconda3/envs/ome/lib/python3.10/site-packages/datashader/compiler.py:302, in <genexpr>(.0)
    300 else:
    301     array_module = np
--> 302 return lambda shape: tuple(c(shape, array_module) for c in creators)
File /mnt/miniconda3/envs/ome/lib/python3.10/site-packages/datashader/reductions.py:785, in by._build_create.<locals>.<lambda>(shape, array_module)
    783 def _build_create(self, required_dshape):
    784     n_cats = len(required_dshape.measure.fields)
--> 785     return lambda shape, array_module: self.reduction._build_create(
    786         required_dshape)(shape + (n_cats,), array_module)
File /mnt/miniconda3/envs/ome/lib/python3.10/site-packages/datashader/reductions.py:472, in Reduction._create_uint32(shape, array_module)
    470 @staticmethod
    471 def _create_uint32(shape, array_module):
--> 472     return array_module.zeros(shape, dtype='u4')
MemoryError: Unable to allocate 74.8 GiB for an array with shape (25783, 35416, 22) and data type uint32
```

Sonja-Stockhaus:
@LucaMarconato  I feel like this also has to do with https://github.com/scverse/spatialdata-plot/issues/296 and might be solved with https://github.com/scverse/spatialdata-plot/pull/309 since with that version of the code everything runs for me.

LucaMarconato:
I just tried: the error disappeared and the plot is fast! There is still a problem with the lims as I get this plot
<img width="771" alt="image" src="https://github.com/user-attachments/assets/5905d143-cc03-401f-bb74-ac29ddf7dcf3">
Can you have a look at it as part of the PR please?

LucaMarconato:
I confirm that my claim seems to be indeed the source of the issue, in fact a workaround that fixes the plot is to call
```python
[...]

PR #309, opened 2024-07-17
datashader speedup and bugfixes

Sonja-Stockhaus:
This speeds everything up for me locally, but for the benchmark (#296), I see an effect (aka datashader faster than matplotlib) for e.g. 10k points/shapes etc.


src/spatialdata_plot/pl/render.py
```diff
@@ -223,16 +238,44 @@ def _render_shapes(
...
+            transformations={"global": Scale([1, factor, factor], ("c", "y", "x"))},
+        )
+
+        # prepare transformation
+        trans = get_transformation(rgba_image, get_all=True)["global"]
+        affine_trans = trans.to_affine_matrix(input_axes=("x", "y"), output_axes=("x", "y"))
+        trans = mtransforms.Affine2D(matrix=affine_trans)
+        trans_data = trans + ax.transData
+
+        rgba_image = np.transpose(rgba_image.data.compute(), (1, 2, 0))  # type: ignore[attr-defined]
```
Sonja-Stockhaus, line 256: Here (and in render_shapes), we access the image as numpy array from the SpatialImage. mypy doesn't believe that `compute()` exists...

src/spatialdata_plot/pl/render.py
```diff
@@ -416,9 +449,15 @@ def _render_points(
                 y_ext = [min(y_ext[0], previous_ylim[1]), max(y_ext[1], previous_ylim[0])]
             else:  # case for e.g. labels
                 y_ext = [min(y_ext[0], previous_ylim[0]), max(y_ext[1], previous_ylim[1])]
```
LucaMarconato, line 451: I'd consider bundling this code in a private function since it's duplicate from above.

src/spatialdata_plot/pl/render.py
```diff
@@ -416,9 +449,15 @@ def _render_points(
...
-        plot_height = int(np.round(y_ext[1] - y_ext[0]))
+
+        # compute canvas size in pixels close to the actual image size to speed up computation
+        plot_width = x_ext[1] - x_ext[0]
+        plot_height = y_ext[1] - y_ext[0]
+        plot_width_px = int(round(fig_params.fig.get_size_inches()[0] * fig_params.fig.dpi))
+        plot_height_px = int(round(fig_params.fig.get_size_inches()[1] * fig_params.fig.dpi))
+        factor = np.min([plot_width / plot_width_px, plot_height / plot_height_px])
+        plot_width = int(np.round(plot_width / factor))
+        plot_height = int(np.round(plot_height / factor))
```
LucaMarconato, line 460: I'd consider bundling this code in a private function since it's duplicate from above.

src/spatialdata_plot/pl/render.py
```diff
@@ -456,8 +495,25 @@ def _render_points(
                 cmap=render_params.cmap_params.cmap,
             )
 
-        rbga_image = np.transpose(ds_result.to_numpy().base, (0, 1, 2))
-        cax = ax.imshow(rbga_image, zorder=render_params.zorder, alpha=render_params.alpha)
+        # create SpatialImage to get it back to original size
```
LucaMarconato, line 498: Also this code is a duplicate, I'd consider it refactoring it into a private function.

LucaMarconato:
Thanks @Sonja-Stockhaus ! I have tried the PR on the Visium HD data and the speed up is significant!

LucaMarconato:
@Sonja-Stockhaus I found a bug with this PR when the rasterized object needs to be aligned with the rest https://github.com/scverse/spatialdata-plot/issues/291. The solution seems straightforward, I think one just needs to initialize the image element from datashader using the old coordinate transformations.

LucaMarconato:
Thanks! Ready for review correct?

timtreis:
Will try to find some time today to deal with

> Sonja, I found a bug with this PR when the rasterized object needs to be aligned with the rest https://github.com/scverse/spatialdata-plot/issues/291. The solution seems straightforward, I think one just needs to initialize the image element from datashader using the old coordinate transformations.

and will ping you afterwards 👍

LucaMarconato:
You can use the code from this branch: https://github.com/scverse/spatialdata-plot/pull/295.

LucaMarconato:
Hi @timtreis, checking the status of this PR. There are still some open tasks on this PR right? Can you list them here as tickable tasks please?

Sonja-Stockhaus:
Todo:

- [ ] fix failing tests
- [ ] look into aggregation methods that aren't working rn (mode, any, std, var)
- [ ] behavior of alpha when rendering shapes should be more similar to mpl
- [ ] outline of polygons need to be implemented in datashader

LucaMarconato:
Thanks for the update!

> outline of polygons need to be implemented in datashader

This task here seems tricky. I would consider raising a warning saying that the outline is currently not supported and workin on this on a separate PR.

Sonja-Stockhaus:
> > outline of polygons need to be implemented in datashader
> 
> This task here seems tricky. I would consider raising a warning saying that the outline is currently not supported and workin on this on a separate PR (with low priority).

If you use `cvs.line` instead of `cvs.polygons` you get only the outline, so you should be able to overlay the two. Depending on how fast we want to merge this PR, we can of course move that to a new one

LucaMarconato:
Ah ok, then it should be a fast addition, thanks for the update.

Sonja-Stockhaus:
General problem of datashader so far: when you render elements, coloring them by a value and a color map, you wouldn't see a single element if all of them have the same value `X.` This is due to the scaling of the data for the colormap that happens in datashader (`scaled_data = (data - span[0])/(span[1] - span[0])`) which leads to all elements being assigned `alpha=0`.

Fix in this PR: internally, not the whole colormap is passed to datashader, but just the color given by `cmap(0.0)`. It is used to color all elements. Still, a colorbar of the cmap is drawn with the range `[X, X+1]`.

LucaMarconato:
Thanks for the explanation, I think the approach that you implemented is a good one!

src/spatialdata_plot/pl/render.py
```diff
@@ -156,27 +158,27 @@ def _render_shapes(
...
-        logger.info(f"Using '{method}' as plotting backend.")
+        logger.info(
+            f"Using '{method}' backend with '{render_params.ds_reduction}' as reduction"
+            " method to speed up plotting. Depending on the reduction method, the value"
+            " range of the plot might change. Set method to 'matplotlib' do disable"
+            " this behaviour."
+        )
 
     if method == "datashader":
         trans = mtransforms.Affine2D(matrix=affine_trans) + ax.transData
```
melonora, line 175: It seems like we could use `trans` defined on line 155 here. Though this line is not changed in this PR perhaps we can change it here.

src/spatialdata_plot/pl/render.py
```diff
@@ -213,26 +230,64 @@ def _render_shapes(
...
-                cmap=render_params.cmap_params.cmap,
+                cmap=ds_cmap,
+                how="linear",
+                min_alpha=np.min([254, render_params.fill_alpha * 255]),
+            )
+
+        # shade outlines if needed
+        outline_color = render_params.outline_params.outline_color
+        if isinstance(outline_color, str) and outline_color.startswith("#") and len(outline_color) == 9:
+            outline_color = outline_color[:-2]
```
melonora, line 266: What if the outline color is given as hex without alpha?

src/spatialdata_plot/pl/render.py
```diff
@@ -193,18 +195,33 @@ def _render_shapes(
...
+                    render_params.ds_reduction, cvs, sdata_filt.shapes[element], col_for_color, "shapes"
+                )
                 # save min and max values for drawing the colorbar
-                aggregate_with_sum = (agg.min(), agg.max())
+                aggregate_with_reduction = (agg.min(), agg.max())
         else:
             agg = cvs.polygons(sdata_filt.shapes[element], geometry="geometry", agg=ds.count())
+        # render outlines if needed
+        render_outlines = render_params.outline_alpha > 0
+        if render_outlines:
```
melonora, line 219:
```suggestion
        if render_outlines := render_params.outline_alpha > 0:
```

src/spatialdata_plot/pl/render.py
```diff
@@ -213,26 +230,64 @@ def _render_shapes(
...
+            rgba_image, trans_data = _create_image_from_datashader_result(ds_outlines, factor, ax)
+            _ax_show_and_transform(
+                rgba_image, trans_data, ax, zorder=render_params.zorder, alpha=render_params.outline_alpha
+            )
+
+        cax = None
+        if aggregate_with_reduction is not None:
             cax = ScalarMappable(
-                norm=matplotlib.colors.Normalize(vmin=aggregate_with_sum[0], vmax=aggregate_with_sum[1]),
+                norm=matplotlib.colors.Normalize(vmin=aggregate_with_reduction[0], vmax=aggregate_with_reduction[1]),
```
melonora, line 290: Does this mean that a `Normalize` object passed on by the user will always be overwritten?

src/spatialdata_plot/pl/utils.py
```diff
@@ -1622,9 +1627,31 @@ def _type_check_params(param_dict: dict[str, Any], element_type: str) -> dict[st
     if param_dict.get("table_name") and not isinstance(param_dict["table_name"], str):
         raise TypeError("Parameter 'table_name' must be a string .")
 
-    if param_dict.get("method") not in ["matplotlib", "datashader", None]:
+    # like this because the following would assign True/False to 'method'
+    # method := param_dict.get("method") not in ["matplotlib", "datashader", None]
+    method = param_dict.get("method")
+    if method not in ["matplotlib", "datashader", None]:
```
melonora, line 1633:
```suggestion
    if (method := param_dict.get("method")) not in ["matplotlib", "datashader", None]:
```

src/spatialdata_plot/pl/utils.py
```diff
@@ -1622,9 +1627,31 @@ def _type_check_params(param_dict: dict[str, Any], element_type: str) -> dict[st
...
+        "count",
+        # "m2", -> not intended to be used alone (see https://datashader.org/api.html#datashader.reductions.m2)
+        # "mode", -> not supported for points (see https://datashader.org/api.html#datashader.reductions.mode)
+        "std",
+        "var",
+        "max",
+        "min",
+    ]
+    ds_reduction = param_dict.get("ds_reduction")
+    if ds_reduction and (ds_reduction not in valid_ds_reduction_methods):
```
melonora, line 1649:
```suggestion

    if (ds_reduction := param_dict.get("ds_reduction")) and (ds_reduction not in valid_ds_reduction_methods):
```

src/spatialdata_plot/pl/utils.py
```diff
@@ -1984,6 +2015,156 @@ def set_zero_in_cmap_to_transparent(cmap: Colormap | str, steps: int | None = No
...
+        rgba_image_data,
+        dims=("c", "y", "x"),
+        transformations={"global": Scale([1, factor, factor], ("c", "y", "x"))},
+    )
+
+    # prepare transformation
+    trans = get_transformation(rgba_image, get_all=True)["global"]
+    affine_trans = trans.to_affine_matrix(input_axes=("x", "y"), output_axes=("x", "y"))
+    trans = mtransforms.Affine2D(matrix=affine_trans)
+    trans_data = trans + ax.transData
```
melonora, line 2070: I see this kind of construct coming back more also in `render.py`. Perhaps refactor in its own separate function `prepare_transformation`?

melonora (changes requested):
Hi @Sonja-Stockhaus, thanks for the amazing work. I think it is very far already, but would like some minor changes here and there, mainly to shorten the code a bit and some refactorings to keep it more maintainable.

src/spatialdata_plot/pl/basic.py
```diff
@@ -419,6 +433,7 @@ def render_points(
                 table_name=param_values["table_name"],
                 zorder=n_steps,
                 method=method,
+                ds_reduction=kwargs.get("datashader_reduction", None),
```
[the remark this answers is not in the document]
Sonja-Stockhaus: Thanks for spotting! As far as I can tell it wouldn't make a difference for the behavior but I agree that it should be consistent

src/spatialdata_plot/pl/render.py
Sonja-Stockhaus: Then the string would have length 7 and therefore the last 2 characters wouldn't be removed. When I tested it, that worked fine :)

src/spatialdata_plot/pl/render.py
Sonja-Stockhaus: That's a good point. I added that part so that the color map has the same range as the actual values. If you remove that I think that a `Normalize` object passed by the user right now has no effect at all (when using datashader), which is not ideal. I'll have a look into that

tests/pl/test_render_shapes.py
```diff
@@ -317,6 +323,44 @@ def test_plot_datashader_can_color_by_value(self, sdata_blobs: SpatialData):
         sdata_blobs.shapes["blobs_polygons"]["value"] = [1, 10, 1, 20, 1]
         sdata_blobs.pl.render_shapes(element="blobs_polygons", color="value", method="datashader").pl.show()
 
+    def test_plot_datashader_can_color_by_identical_value(self, sdata_blobs: SpatialData):
```
timtreis, line 326: Please elaborate why we need this

tests/pl/test_render_shapes.py
```diff
@@ -317,6 +323,44 @@ def test_plot_datashader_can_color_by_value(self, sdata_blobs: SpatialData):
...
+        ).pl.show()
+
+    def test_plot_datashader_can_render_with_rgb_colored_outline(self, sdata_blobs: SpatialData):
+        sdata_blobs.pl.render_shapes(
+            method="datashader", element="blobs_polygons", outline_alpha=1, outline_color=(0.0, 0.0, 1.0)
+        ).pl.show()
+
+    def test_plot_datashader_can_render_with_rgba_colored_outline(self, sdata_blobs: SpatialData):
+        sdata_blobs.pl.render_shapes(
+            method="datashader", element="blobs_polygons", outline_alpha=1, outline_color=(0.0, 1.0, 0.0, 1.0)
```
timtreis, line 361: Throw warning that outline_alpha/outline_fill take precedent over rgbA, and we'll discard the A then.

timtreis (changes requested):
as discussed

src/spatialdata_plot/pl/render.py
Sonja-Stockhaus: Passing a `Normalize` object should work now. Not sure about the desired behavior of `clip` (see matplotlib example #372). @timtreis and I just discussed that we need more tests for this anyways, so I'm leaving it as is and we can do a follow-up

Sonja-Stockhaus:
As discussed with @timtreis, we now use `sum` as default datashader aggregation for points and `mean` for shapes

PR #323, opened 2024-08-15
Categorical data is now correctly colored in the legend when plotting v2

timtreis:
Successor of https://github.com/scverse/spatialdata-plot/pull/262 to avoid solving unneccecary merge conflicts due to age


LucaMarconato:
Amazing! Great work!

src/spatialdata_plot/pl/render.py
```diff
@@ -774,17 +794,19 @@ def _render_labels(
         table_name=table_name,
     )
 
-    if (render_params.fill_alpha != render_params.outline_alpha) and render_params.contour_px is not None:
-        # First get the labels infill and plot them
+    # default case: no contour, just fill
+    # if fill_alpha and outline_alpha are the same, we're technically also at a no-outline situation
```
Sonja-Stockhaus, line 798: @timtreis Probably just an edge case, but if the user increases the outline width when `fill_alpha == outline_alpha` one would expect the labels to get bigger which would not happen since it's treated as no-outline, correct?

timtreis: hm, yes, that's true.

PR #334, opened 2024-08-28
Donut-MultiPolygons are now correctly rendered again

timtreis:
The extent of the circle is not "perfect", but that's because back then when we first made the plot, we were still doing the extent-calculations "in-house". Now they've been ported upstream so we have no control over them anymore. 
image

(also deleting two unused functions)


src/spatialdata_plot/pl/utils.py
```diff
@@ -10,11 +10,11 @@
 from typing import Any, Literal, Union
 
 import matplotlib
+import matplotlib.patches
 import matplotlib.patches as mpatches
 import matplotlib.patches as mplp
```
Sonja-Stockhaus, line 15: Any particular reason why we need all 3 lines?

Sonja-Stockhaus (approved):

src/spatialdata_plot/pl/utils.py
timtreis: No 🥸

Issue #324, opened 2024-08-16
Unable to set vmin vmax when plotting vector data


wangjiawen2013:
Hi,
I want set the vmin and vmax of the color, but failed. This is the first method:
```
norm = matplotlib.colors.Normalize(vmin=-1.0, vmax=0.4)
sdata.pl.render_shapes("cell_circles", norm=norm, color="NOTCH2", method="matplotlib").pl.show(coordinate_systems= "physical", figsize=(10,4))
```
The resulting plot is here:
![image](https://github.com/user-attachments/assets/ebea5ffd-45bd-4f9d-9b82-4dbb91388dd1)

This is the second method:
`sdata.pl.render_shapes("cell_circles", vmin=-1, vmax=1, color="NOTCH2", method="matplotlib").pl.show(coordinate_systems= "physical", figsize=(10,4))`
The resulting plot is here:
![image](https://github.com/user-attachments/assets/b4668608-eb7e-40b7-9ba9-38310e22324d)

`vmin` and `vmax` didn't work in both cases.

LucaMarconato:
Hi, this is related to this https://github.com/scverse/spatialdata-plot/issues/303, please try using `clip=True`.

wangjiawen2013:
@LucaMarconato 
`clip` still doesn't work, Could you show me an example on how to set colorbar vmin and vmax ?

LucaMarconato:
Thanks for the observation. I have changed the title to emphasize that the issue is also open for vector data (the linked issue is indeed for raster data).
Regarding the missing colorbar, I already reported the issue here: https://github.com/scverse/spatialdata-plot/issues/304, so we can follow up there. The developer assigned to the issue is temporarily not available for the next 14 days because of a pressing deadline.

olivermccallion:
Running into the same problem with spatialdata-plot version 0.2.6.dev3+gffd4a1f where:

```
region.pl.render_shapes(
    "cell_boundaries", color = "CD3E", vmax = 11, clip=True).pl.show(title="CD3", coordinate_systems="global", ax=axes[1])
```
```
region.pl.render_shapes(
    "cell_boundaries", color = "CD3E", vmax = 11).pl.show(title="CD3", coordinate_systems="global", ax=axes[1])
```
and
```
norm = matplotlib.colors.Normalize(vmin=0.0, vmax=2000)
region.pl.render_shapes(
[...]

PR #344, opened 2024-09-03
Coloring labels by a continuous variable fixed

timtreis:
As reported by @clwgg, there was a bug with the color of continuous annotations which resulted in them having incorrect colors. This is now fixed and the tests have been updated. Furthermore, within this fix I noticed our testing threshold was way too high and let a bunch of bugs pass. So I also lowered that and fixed the bugs this revealed.


timtreis:
The two failing tests can probably be traced back to https://github.com/scverse/spatialdata-plot/issues/324

src/spatialdata_plot/pl/utils.py
```diff
@@ -808,44 +800,43 @@ def _map_color_seg(
...
     cell_id = np.array(cell_id)
-    if color_vector is not None and isinstance(color_vector.dtype, pd.CategoricalDtype):
-        # users wants to plot a categorical column
+
+    if pd.api.types.is_categorical_dtype(color_vector.dtype):
+        # Case A: users wants to plot a categorical column
         if np.any(color_source_vector.isna()):
             cell_id[color_source_vector.isna()] = 0
-        val_im: ArrayLike = map_array(seg, cell_id, color_vector.codes + 1)
+        val_im: ArrayLike = map_array(seg.copy(), cell_id, color_vector.codes + 1)
```
melonora, line 808: hmm is the copy really required here? I thought that map array returns an array with same shape as the original array, but it would not have the same pointer

timtreis: Ah, I remember. Don't ask me why, but if I don't `copy()`, one of the tests fails with:

`ValueError: buffer source array is read-only`

clwgg: I had the same issue with the current `main`: whenever I tried to color labels by a column in `table.obs`, I'd get `ValueError: buffer source array is read-only`, and putting in those `.copy()` calls fixed it. This happened regardless of whether the `SpatialData` object was buffered to disk or not.

melonora: I see, this is probably a bug in skimage actually. For now we could keep it as it is, but please add a comment with the issue so we remove this when we can.

melonora (approved):
preapproving for the vmin vmax fix, but please if you can open an issue in skimage regarding the buffer error. I have seen similar errors in pandas at some point, I believe it had to do with some cython code not allowing for read only buffer, but this could be fixed I think and then prevents a copy.

PR #346, opened 2024-09-04
Refactor of `colorbar` and `norm` logic

timtreis:
Unified `norm` logic across the 4 functions. Before, the `render_images` function would take a separate `percentiles_for_norm` argument (legacy reasons) which would/could interfere with whatever was passed to `norm`. As a consequence of this, I fixed some of the other resulting images.


CHANGELOG.md
```diff
@@ -8,6 +8,26 @@ and this project adheres to [Semantic Versioning][].
...
+### Added
+
+-
+
+### Changed
+
+-   Lowered RMSE-threshold for plot-based tests from 45 to 15 (#344)
+-   When subsetting to `groups`, `NA` isn't automatically added to legend (#344)
+-   When rendering a single image channel, a colorbar is now shown (#346)
+-   Removed `quantiles_for_norm` parameter (#346)
```
melonora, line 22:
```suggestion
-   Removed `percentiles_for_norm` parameter (#346)
```

melonora (approved):
Seems good to me, though please check my comment also in #344 regarding the skimage issue. Also please add a PR description before merging with also the reasoning of getting rid of percentiles_for_norm:)

clwgg:
Apologies for chiming in here again, but I'm curious about the decision to remove `percentiles_for_norm`, and I think it would be helpful to get some pointers on how its functionality should be replaced once I pull in these changes. I've found its capacity to do per-channel normalization quite helpful when there are extreme differences in the intensities of different markers. For example, using this:

``` python
sdata.pl.render_images(
    '78_image',
    channel=["CD45", "PanCK", "DAPI"]
).pl.show(coordinate_systems='78')
```

I'd get an image like this:


Because of data like that, I've routinely started plotting like this:
``` python
sdata.pl.render_images(
    '78_image',
    percentiles_for_norm=(0, 100),
    channel=["CD45", "PanCK", "DAPI"]
).pl.show(coordinate_systems='78')
```

which for this example gives:


How would you achieve something like this using just the `norm` flag?

timtreis:
Hm, interesting. Does passing a matplotlib `Normalise` achieve a different outcome?

timtreis:
One caveat that I see in that, this way, all channels are normalised based on the same vmin/vmax which might not necessarily work if your channels have different value ranges. We have a mostly undocumented `transfunc` parameter in one of the functions (currently on the go, will check later).

Is the goal to scale all channels to [0,1]? I'd be happy to build out that parameter for this purpose

clwgg:
yeah exactly, I think having some control over which channels get which vmin/vmax would be super helpful. Even with `percentiles_for_norm` I've sometimes wondered how it would look if I could set a list of `(pmin, pmax)` tuples with the same length as `channel`, to have some more fine-grained control over which channels to highlight. I suppose something similar could be achieved if it was possible to set a list of `Normalize` objects with the same length as channel. But achieving the task of `percentiles_for_norm(0, 100)` (which I think would still be the most common use-case for me) would get a lot more wordy since I'd first have to figure out the per-channel vmin/vmax corresponding to each channels min and max, and instantiate the appropriate `Normalize` objects to hand off to `render_images`.

PR #349, opened 2024-09-04
Colorbar is now the same height as the `ax` object

timtreis:


giovp:
magic 🤩

LucaMarconato:
Tasks missing to close this PR:
- [ ] right now, the colobar gets defined in 2 places in the codebase -> unify

Additional task, suitable for the Basel Hackathon:
- [ ] establish set of parameters relevant for cbar: distance from plot (probably just scalar that multiplied a fixed), cbar width, cbar title

PR #393, opened 2024-12-17
fix isinstance

LucaMarconato:
Small fix. I while ago I started getting an error here:

```python
if isinstance(colors[params_copy.color].dtype, pd.CategoricalDtype):
```

because `colors[params_copy.color].dtype` was a dataframe instead of a series, and therefore `dtype` was not present. I lost the code I used to reproduce this because I had put together a solution quickly while being busy in another PR.

The behavior is due to the fact that `colors = sc.get.obs_df(some_adata, some_keys)` is a `pd.DataFrame` if `some_keys` is an `Iterable`, while it is a `pd.Series` when `some_keys` is a `str` (this is how `sc.get.obs_df()` behaves).


LucaMarconato:
When tests pass I am going to update the changelog.

src/spatialdata_plot/pl/basic.py
```diff
@@ -950,8 +950,10 @@ def show(
 
                     if wanted_labels_on_this_cs:
                         if (table := params_copy.table_name) is not None:
-                            colors = sc.get.obs_df(sdata[table], params_copy.color)
-                            if isinstance(colors[params_copy.color].dtype, pd.CategoricalDtype):
+                            keys = [params_copy.color] if not isinstance(params_copy.color, list) else params_copy.color
```
LucaMarconato, line 953: I got lost on the type of `params_copy.color`. Is this always a `str`? Or is it sometimes an `Iterable`? Can it be of `len() > 1` here? The tests pass but if we know it's always a string we can write better code.

LucaMarconato:
@timtreis @Sonja-Stockhaus @melonora very small PR that fixes some bugs I got locally a while ago, can you please have a look?

LucaMarconato:
Test pass, so for the sake of a release I'd be inclined to merge soon.

src/spatialdata_plot/pl/basic.py
Sonja-Stockhaus: I'm pretty sure it can never be an Iterable, but would like to check more carefully before really relying on this

Sonja-Stockhaus (approved):
Lgtm, if you want to merge soon @LucaMarconato, we can check the str/Iterable thing in a later PR

src/spatialdata_plot/pl/basic.py
melonora: it can never be an interable, it is always either color like or it corresponds to a particular column that then contains the color.

melonora (changes requested):
Can you please provide first a small reproducible example that shows the issue is still there on main because I think I already implemented a fix for this a while back. Main reason was that the scanpy function being used had a different return type. Not too certain about it but I would like to see an example first.

LucaMarconato:
Thanks for the review, I will try to find the plot that was causing the problem.

LucaMarconato:
Ok I looked more into this. What caused the problem is due to `scanpy==1.10.3`, now fixed in `1.10.4`. Shown is the diff between `scanpy` `1.10.3` and `1.10.4`. The left version leads to a `pd.DataFrame` when the argument to `get.obs_df()` is a `list[str]`, while it leads to a `pd.Series` when a `str` is passed. The second always passes a `list[str]`, so the return is consistently a `pd.DataFrame`.

We could also fix this by requiring `1.10.4` as a min version. I'll think about it and update this PR.
image

LucaMarconato:
I tested it against `scanpy` `1.10.3` and `1.10.4` and the problem is addressed.

LucaMarconato:
Merging since the problem is now addressed.

PR #397, opened 2024-12-20
Update contributing.md

LucaMarconato:
FYI, this is the command that I use for `act`:

```bash
echo '\033[0;31mremove -r when testing a new branch\033[0m' && time act test --artifact-server-path ~/temp/act --env CODECOV_TOKEN=secret_token_you_need_to_generate --container-architecture linux/arm64 -r -j test 
```

Please notice that you have to change `arm64` to `amd64` if you are not using Apple Silicon.
It may work even without the `codecov` token, please ask me if you encounter problem and I'll help with that.

PR #413, opened 2025-01-28
fix table name and add unit test for colors passed via .uns

ArneDefauw:
`spatialdata_plot.pl.render._render_labels`  still had `sdata.table` , changed to `sdata.tables[table_name]`, otherwise colors passed via .uns where ignored.

also in https://github.com/scverse/spatialdata-plot/blob/43f25f654e123ca3a6a5da0e499ef15f22d0ec27/src/spatialdata_plot/pl/utils.py#L854

I would remove the check `cluster_key` in `adata.uns`. Because of this I had to set

        `sdata_blobs[ "other_table" ].uns[ "category" ] = "__value__"`

as a placeholder in https://github.com/ArneDefauw/spatialdata-plot/blob/b5cb0260c32a859783f87f10b8af05dc26feab76/tests/pl/test_render_labels.py#L219 . 

I also stumbled upon a subtle bug while testing this feature. With the small fix in this PR, one can pass colors via .uns[ f"{cluster_key}_colors" ], and all works as expected, see 
https://github.com/ArneDefauw/spatialdata-plot/blob/b5cb0260c32a859783f87f10b8af05dc26feab76/tests/pl/test_render_labels.py#L219

We get:
no_query

But if we do a bounding box query, we get:
query_fail

While we expect:
Figure_1

In short, this is caused by https://github.com/scverse/spatialdata/blob/03d3be80fad69ff54097e90a9e80ad02e9e0e242/src/spatialdata/_utils.py#L203 

also see  https://github.com/scverse/anndata/issues/997 

For details see https://github.com/ArneDefauw/spatialdata-plot/blob/b5cb0260c32a859783f87f10b8af05dc26feab76/tests/pl/test_render_labels.py#L227 and possible workaround https://github.com/ArneDefauw/spatialdata-plot/blob/b5cb0260c32a859783f87f10b8af05dc26feab76/tests/pl/test_render_labels.py#L252

Note that being able to plot a subset in the same colors as the orignal is a rather common use case. However, I am not sure where to fix. Maybe documenting the workaround suffices for now.

### DIFF
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -755,7 +755,7 @@ def _set_color_source_vec(
         color_source_vector = pd.Categorical(color_source_vector)  # convert, e.g., `pd.Series`
 
         color_mapping = _get_categorical_color_mapping(
-            adata=sdata.table,
+            adata=sdata.tables[table_name],
             cluster_key=value_to_plot,
             color_source_vector=color_source_vector,
             groups=groups,
--- a/tests/pl/test_render_labels.py
+++ b/tests/pl/test_render_labels.py
@@ -10,13 +10,15 @@ from spatial_image import to_spatial_image
 from spatialdata import SpatialData, deepcopy, get_element_instances
 from spatialdata.models import TableModel
 
+from spatialdata import bounding_box_query
+
 import spatialdata_plot  # noqa: F401
 from tests.conftest import DPI, PlotTester, PlotTesterMeta
 
 RNG = np.random.default_rng(seed=42)
 sc.pl.set_rcParams_defaults()
 sc.set_figure_params(dpi=DPI, color_map="viridis")
-matplotlib.use("agg")  # same as GitHub action runner
+#matplotlib.use("agg")  # same as GitHub action runner
 _ = spatialdata_plot
 
 # WARNING:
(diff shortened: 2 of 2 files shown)


LucaMarconato:
Hi @ArneDefauw, thanks for the contribution. Two initial comments:
- could you please open one issue describing what the PR addresses? This makes it easier to review and keeps a cleaner history of what a PR introduces. At the moment it's not immediate and one has to go and check the code and discussion. Or in alternative, without having to open an issue, can you please add a summary at the beginning of the PR of the issues being addressed? Thanks a lot!
- I see some tests are not passing, one seems to be due to a difference between an old plot and the new one, while other seems to be due to raised exception. Do you have fix for them or comments on them?

Thanks!

ArneDefauw:
Hi @LucaMarconato ,

I've linked an issue.

For the failed unit tests, there was a KeyError, because I overlooked that `table_name` can be `None`. I changed it in the PR.

For me all unit tests fail locally on the main branch of spatialdata-plot, so it is difficult to test. Therfore, I did not include base images for the unit tests I've added, because they would probably fail anyway in the CI/CD testing environment. 

The unit tests I've added are for documentation of the issue, some of them can probably be removed when merging to main.

LucaMarconato:
Thanks for updating the PR! Regarding the failing tests, could you please give a try to the approach described here? https://github.com/scverse/spatialdata-plot/blob/main/docs/contributing.md

Locally generally the tests are reproducible on Ubuntu, but for other OS I use a Docker image, as described in the link above.

LucaMarconato:
My configuration with PyCharm is described here https://github.com/scverse/spatialdata-plot/pull/397.

ArneDefauw:
I managed to fix the tests (using a centOS server). 

Note that I had to move this line https://github.com/scverse/spatialdata-plot/blob/478ab27f9e8a6e06fc28c411e7721f1332b34b66/tests/pl/test_render_labels.py#L16 inside https://github.com/scverse/spatialdata-plot/blob/478ab27f9e8a6e06fc28c411e7721f1332b34b66/tests/pl/test_render_labels.py#L217, respectively https://github.com/scverse/spatialdata-plot/blob/478ab27f9e8a6e06fc28c411e7721f1332b34b66/tests/pl/test_render_labels.py#L237

to have the same results when running one test versus all the tests, because its internal state was moving forward due to multiple calls to it. Because of this I also had to update this figure https://github.com/scverse/spatialdata-plot/blob/main/tests/_images/Labels_can_annotate_labels_with_table_layer.png

LucaMarconato:
Fantastic! Thanks a lot!

Issue #418, opened 2025-02-04
`norm` doesn't have impact for single image channel rendering when `clip=False`


Sonja-Stockhaus:
```python
from matplotlib.colors import Normalize
from matplotlib import rcParams
import spatialdata_plot
from spatialdata.datasets import blobs

blob=blobs()

cmap = matplotlib.colormaps[rcParams["image.cmap"]]
cmap.set_under("black")
cmap.set_over("grey")

norm = Normalize(vmin=0.3, vmax=0.7, clip=False)

blob.pl.render_images(element="blobs_image", channel=0, norm=norm, cmap=cmap).pl.show()
``` 
The result looks like the original image when you render channel 0 without any `norm`, except for the colorbar:

![Image](https://github.com/user-attachments/assets/80da9c29-f74b-4037-80e6-b87cddc6f186)

This is because `imshow()` gets the already normalized data, but doesn't know that - so it autoranges to show everything on the colormap.
When passing the `Normalize` object directly to the `norm` argument of `imshow`, we get the desired behavior:

![Image](https://github.com/user-attachments/assets/3d7ff12f-3ef5-4874-84ad-b2e69bd743ca)

Issue #372, opened 2024-10-16
Missing tests for norm=Normalize(...)


Sonja-Stockhaus:
Currently, we don't have any tests including the `norm=Normalize(...)` argument. It should work for matplotlib and also datshader (after #309), but I think we need some tests and maybe also a bit more documentation.

Also, I noted that both seem to give the same result, not sure if that's a bug
```python
from spatialdata.datasets import blobs
from matplotlib.colors import Normalize
blob=blobs()

blob.pl.render_points(color="instance_id", size=40, norm=Normalize(4, 7, clip=True)).pl.show()
blob.pl.render_points(color="instance_id", size=40, norm=Normalize(4, 7, clip=False)).pl.show()
``` 
![Image](https://github.com/user-attachments/assets/7afb76af-c0ce-42bf-9ee9-af6870f25a97)
![Image](https://github.com/user-attachments/assets/53e579b9-1c6a-4fa4-b8dc-f4aeb7063e20)

LucaMarconato:
Thanks for reporting. The `instance_id` column has values in `[0, 1, ..., 9]`. With `clip=True` one would expect the values `[0, 1, 2, 3, 4]` to have the same color, and the values `[7, 8, 9]` to have the same color. So it seems like that there is a bug.

Sonja-Stockhaus:
When using `datashader`, both plots look identical as well.

Sonja-Stockhaus:
After looking into this a bit more, I would say the behavior is not a bug: according to the `matplotlib` documentation (https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html), with `clip=False`, values outside of `[vmin, vmax]` are mapped to values outside `[0, 1]` (and not exactly to 0 or 1 as would happen with `clip=True`). Then, the respective colormap can "can mark these under and over values with specific colors".
In the examples above, we use the default colormap `viridis`, which uses its minimal value for _under_ values and its maximal value for _over_ values (see below).

![Image](https://github.com/user-attachments/assets/e07170b5-e227-42d4-8137-adadc09948b3)

So, if you would set _over_ and _under_ differently, the result looks like this for `clip=False` (the `clip=True` plot looks like above)
```python
from matplotlib import rcParams
import matplotlib

cmap = matplotlib.colormaps[rcParams["image.cmap"]]
cmap.set_under("grey")
cmap.set_over("grey")

[...]

PR #419, opened 2025-02-04
Fix norm behavior, add tests

Sonja-Stockhaus:
should solve https://github.com/scverse/spatialdata-plot/issues/418 and https://github.com/scverse/spatialdata-plot/issues/372


src/spatialdata_plot/pl/render.py
```diff
@@ -794,7 +804,9 @@ def _render_images(
         cmap._init()
         cmap._lut[:, -1] = render_params.alpha
 
-        _ax_show_and_transform(layer, trans_data, ax, cmap=cmap, zorder=render_params.zorder)
+        _ax_show_and_transform(
+            layer, trans_data, ax, cmap=cmap, zorder=render_params.zorder, norm=render_params.cmap_params.norm
+        )
```
Sonja-Stockhaus, line 809: comment explanation why we pass norm here

tests/pl/test_render_shapes.py
```diff
@@ -456,6 +456,86 @@ def test_plot_can_do_non_matching_table(self, sdata_blobs: SpatialData):
 
         sdata_blobs.pl.render_shapes("blobs_circles", color="instance_id").pl.show()
 
+    def test_plot_can_color_with_norm_no_clipping(self, sdata_blobs: SpatialData):
+        blob = deepcopy(sdata_blobs)
+        blob["table"].obs["region"] = "blobs_polygons"
```
timtreis, line 461: You create the same data obj in all tests, please move to conftest and get it from there

tests/pl/test_render_points.py
```diff
@@ -225,6 +226,72 @@ def test_plot_datashader_can_transform_points(self, sdata_blobs: SpatialData):
...
+            size=40,
+            norm=Normalize(5, 5, clip=True),
+            cmap=cmap,
+            method="datashader",
+            datashader_reduction="max",
+        ).pl.show()
+
+    def test_plot_datashader_norm_vmin_eq_vmax_without_clip(self, sdata_blobs: SpatialData):
+        cmap = matplotlib.colormaps["viridis"]
+        cmap.set_under("black")
```
timtreis, line 284: you always create the same cmap here. Please make some primitive helper function like `_get_cmap_with_under_over("viridis", "black", "grey")` which you can then also reuse in your conftest

src/spatialdata_plot/pl/utils.py
```diff
@@ -2266,3 +2269,51 @@ def _get_transformation_matrix_for_datashader(
...
+    clip: bool = True,
+) -> ds.tf.Image | np.ndarray[Any, np.dtype[np.uint8]]:
+    """ds.tf.shade() part, ensuring correct clipping behavior.
+
+    If necessary (norm.clip=False), split shading in 3 parts and in the end, stack results.
+    This ensures the correct clipping behavior, because else datashader would always automatically clip.
+    """
+    if not clip and isinstance(cmap, Colormap) and span is not None:
+        # in case we use datashader together with a Normalize object where clip=False
+        # why we need this is documented in https://github.com/scverse/spatialdata-plot/issues/372
```
timtreis, line 2289: good!

src/spatialdata_plot/pl/utils.py
```diff
@@ -2266,3 +2269,51 @@ def _get_transformation_matrix_for_datashader(
             tm = tm @ _get_datashader_trans_matrix_of_single_element(x)
         return tm
     return _get_datashader_trans_matrix_of_single_element(trans)
+
+
+def _datashader_shade(
```
timtreis, line 2274: please pick a better, more descriptive, name. Only intuitive for people that know what datashader does

src/spatialdata_plot/pl/render.py
```diff
@@ -656,8 +663,14 @@ def _render_points(
             vmin = aggregate_with_reduction[0].values if norm.vmin is None else norm.vmin
             vmax = aggregate_with_reduction[1].values if norm.vmax is None else norm.vmax
             if (norm.vmin is not None or norm.vmax is not None) and norm.vmin == norm.vmax:
-                vmin = norm.vmin
-                vmax = norm.vmin + 1
+                vmin = norm.vmin - 0.5
+                vmax = norm.vmin + 0.5
+                # if norm.clip:
+                #     vmin = norm.vmin
+                #     vmax = norm.vmin + 1
```
timtreis, line 670: comment.

Also, why the change?

src/spatialdata_plot/pl/render.py
```diff
@@ -631,14 +636,16 @@ def _render_points(
...
                 aggregate_with_reduction = (aggregate_with_reduction[0], aggregate_with_reduction[0] + 1)
 
-            ds_result = ds.tf.shade(
+            ds_result = _datashader_shade(
                 agg,
                 cmap=ds_cmap,
-                how="linear",
+                span=ds_span,
+                clip=norm.clip,
+                min_alpha=np.min([254, render_params.alpha * 255]),
```
timtreis, line 648: I'd be confused by this in 6 months (already am tbh), could you add a comment why we use these super specific numbers?

timtreis (changes requested):

src/spatialdata_plot/pl/render.py
Sonja-Stockhaus: I thought it was more intuitive if both versions (clip=True and clip=False) work the same way, by having the value in the middle of the colormap. The lower end of the colormap would be problematic for clip=False when the low and high values of the cmap are identical to cmap(0) and cmap(1), because then, over values would be mapped to cmap(1) and the actual value (norm.vmin = norm.vmax) to the same color as the under values, which is cmap(0).

src/spatialdata_plot/pl/render.py
Sonja-Stockhaus: That's a good point, because I remember having a reason for doing this np.min thing one year ago... Just not which one exactly XD. The 255 is because datashader wants to get alpha values in a range of [0, 255]. But I don't see a reason (anymore) why we would want to keep min_alpha at max. 254. Maybe that was from when I was still using the alpha and min_alpha parameters at the same time. I'll try to refactor that and see if any test fails

Sonja-Stockhaus: e.g. when you plot using `min_alpha=render_params.alpha * 255` instead of the np.min() thing, 
```python
blob.pl.render_points(color="genes", groups="gene_b", palette="lightgreen", size=20, method="datashader").pl.show()
``` 
will give you this:
grafik
 Instead of
grafik

Apparently this is how I solved that issue last year, maybe there's a more elegant solution

timtreis: I'm fine with your solution, but just copying my comment from the other thing

> Agree. Can you add a comment explaining this? Every random number that suddenly pops up somewhere always preemptively gives me PTSD of maybe having to trace back why it exists 😅

timtreis (approved):
lgtm

PR #434, opened 2025-03-25
Modernize setup

Zethson:
There's a lot to be done but that only took me a few minutes.

- Use Ruff format instead of Black
- Use UV for RTD significantly speeding it up
- Remove flake8 and other old things
- Remove preprocessing from the docs (which is what I actually only wanted to do)


timtreis (approved):
merci!

Issue #443, opened 2025-04-08
plot continous column render labels


ArneDefauw:
Rasterizing a labels layer during plotting, can result in removal of small labels, resulting in inconsitenties.

Minimal example to reproduce the issue

```
import numpy as np
import pandas as pd
from spatialdata.datasets import blobs
from spatialdata.models import TableModel

import spatialdata_plot

RNG = np.random.default_rng(seed=42)

labels_name = "blobs_labels"

sdata_blobs = blobs()

from spatialdata.models import Labels2DModel
import dask.array as da

labels = sdata_blobs[ "blobs_labels" ].data.compute()

# make label 1 small

mask = ( labels == 1 )
labels[ mask ] = 0
labels[ 200, 200  ] = 1

sdata_blobs[ "blobs_labels" ] = Labels2DModel.parse(labels )
```

If we now do 

`sdata_blobs.pl.render_labels( "blobs_labels", color="channel_0_sum", table_name="table" ).pl.show()`

we get:

![Image](https://github.com/user-attachments/assets/3419a1aa-25ca-44f8-9312-60b27d81fb32)

which is correct. However, if we do:

```

arr=da.tile( sdata_blobs[ "blobs_labels" ], ( 4,4 ) )
sdata_blobs[ "blobs_labels_large" ] = Labels2DModel.parse(arr)
adata=sdata_blobs[ "table" ]

adata.obs[ "region" ] = "blobs_labels_large"

adata.uns.pop(TableModel.ATTRS_KEY)

adata = spatialdata.models.TableModel.parse(
    adata,
    region_key="region",
    region="blobs_labels_large",
    instance_key="instance_id",
)
sdata_blobs.pl.render_labels( "blobs_labels_large", color="channel_0_sum", table_name="table" ).pl.show()
```

we get:

![Image](https://github.com/user-attachments/assets/b90c4726-7c6b-470c-848f-1e0857a42867)

while we expect:

![Image](https://github.com/user-attachments/assets/bcc64471-0af4-424f-81cd-46086c21f285)

PR #444, opened 2025-04-08
plot continous column render labels

ArneDefauw:
Fix for https://github.com/scverse/spatialdata-plot/issues/443


LucaMarconato:
Thanks for the fix @ArneDefauw, works great! I'll add my review comments.

LucaMarconato:
I added tests for the continuous case and for the categorical case (see below): 
Screenshot 2025-04-26 at 16 04 08
Labels_can_handle_dropping_small_labels_after_rasterize_categorical

There is still one line that is not covered by tests, I will add a comment in the code.

LucaMarconato:
I uploaded the artifacts from my usual conda env instead of the Docker env I use to simulate the GitHub Actions. I will reupload.

LucaMarconato (approved):
Thanks again!

PR #446, opened 2025-04-16
Get one shape transform

melonora:
closes: #447 

In render shapes, when using datashader, all transforms were obtained instead of the transform to the coordinate system which is supposed to be rendered. This was a problem for the visium notebook when trying the following code snippet:

```
import spatialdata as sd
import spatialdata_plot  

sdata = sd.read_zarr("C:\\Users\\w-mv\\PycharmProjects\\spatialdata-notebooks\\notebooks\\examples\\visium_hd.zarr")

plt.figure(figsize=(10, 10))
ax = plt.gca()

gene_name = "AA986860"
sdata.pl.render_shapes("Visium_HD_Mouse_Small_Intestine_square_016um", color=gene_name, method="datashader").pl.show(
    coordinate_systems="Visium_HD_Mouse_Small_Intestine", ax=ax
)
```

After fixing that, the transform was applied and again parsed by `ShapesModel`. This ensured that after that one can only have a `shape` with coordinate system `global`. This means that the following should always use coordinate system `global`:
```
plot_width, plot_height, x_ext, y_ext, factor = _get_extent_and_range_for_datashader_canvas(
            transformed_element, "global", ax, fig_params
        )
```

### DIFF
--- a/src/spatialdata_plot/pl/render.py
+++ b/src/spatialdata_plot/pl/render.py
@@ -185,7 +185,7 @@ def _render_shapes(
             sdata_filt.shapes[element].loc[is_point, "geometry"] = _geometry[is_point].buffer(scale.to_numpy())
 
         # apply transformations to the individual points
-        element_trans = get_transformation(sdata_filt.shapes[element])
+        element_trans = get_transformation(sdata_filt.shapes[element], to_coordinate_system=coordinate_system)
         tm = _get_transformation_matrix_for_datashader(element_trans)
         transformed_element = sdata_filt.shapes[element].transform(
             lambda x: (np.hstack([x, np.ones((x.shape[0], 1))]) @ tm)[:, :2]
@@ -195,7 +195,7 @@ def _render_shapes(
         )
 
         plot_width, plot_height, x_ext, y_ext, factor = _get_extent_and_range_for_datashader_canvas(
-            transformed_element, coordinate_system, ax, fig_params
+            transformed_element, "global", ax, fig_params
         )
 
         cvs = ds.Canvas(plot_width=plot_width, plot_height=plot_height, x_range=x_ext, y_range=y_ext)


src/spatialdata_plot/pl/render.py
```diff
@@ -195,7 +195,7 @@ def _render_shapes(
         )
 
         plot_width, plot_height, x_ext, y_ext, factor = _get_extent_and_range_for_datashader_canvas(
-            transformed_element, coordinate_system, ax, fig_params
+            transformed_element, "global", ax, fig_params
```
LucaMarconato, line 198: @melonora why `"global"` here? What if `"global"` is not used?

timtreis: @melonora why was this merged? I think Lucas concern is valid here

melonora: Already talked with Luca (sorry tagging does not work on github app) and it is due to the recent change in which global is not required. This caused errors for datasets not having global. However, here the element first gets transformed and after transform the default cs in case nothing is passed on is global. Hence, second call always uses global.

timtreis: But shouldn't we preserve whatever name the user gave the cs?

melonora: I will have another look when I am at my laptop again. Note that the 'coordinate_system' parameter itself is not being altered here. The point is that after the transform there is no other cs but global.

timtreis: At the very end of basic.py we're assigning these as titles here: https://github.com/scverse/spatialdata-plot/blob/7716d438cf5018316ce7d7b2296e5a2e8989403c/src/spatialdata_plot/pl/basic.py#L995

melonora: just as a last follow up, this is not an issue as the title is still correspondent to the coordinate system used. I would have to follow up but it could be that the coordinate system argument in the second call where now the value by default is "global" can be deprecated. 

For now just as final proof of the visium hd example in the notebooks. The following code snippet before was failing:
```
import spatialdata as sd
import spatialdata_plot

sdata = sd.read_zarr("visium_hd.zarr")

plt.figure(figsize=(10, 10))
ax = plt.gca()

gene_name = "AA986860"
sdata.pl.render_shapes("Visium_HD_Mouse_Small_Intestine_square_016um", color=gene_name, method="datashader").pl.show(
    coordinate_systems="Visium_HD_Mouse_Small_Intestine", ax=ax
)
```

However, this PR fixes the issue of having the wrong coordinate system as argument in the second call and gives you this plot:
image

PR #451, opened 2025-04-22
Fix multi-channel compositing and add per-channel norm support

timtreis:
## Summary

- Signal-based multi-channel compositing: replaces 4 inconsistent compositing paths in `_render_images` with a shared `_additive_blend` helper. For each channel, the "signal" (deviation from `cmap(0)`) is summed onto a canvas colored by the mean zero-value. This handles both Napari-style additive blending (black-to-color LUTs) and ImageJ Composite Invert (white-to-color LUTs), and produces reasonable results for arbitrary colormaps.
- Per-channel normalization: `render_images(norm=...)` now accepts a list of `Normalize` objects for independent contrast control per channel (e.g. bright DAPI vs dim cytokine stain).
- Removes dead `percentiles_for_norm` field and its deprecation alias.

## Bug fixes

| Bug | Fix |
|-----|-----|
| Averaging instead of additive sum in multi-cmap paths | Signal-based blend via `_additive_blend` |
| Missing clip to [0, 1] in 2-3 channel path | `np.clip` in `_additive_blend` |
| Double-alpha application in 4+ channel path (baked into RGBA and passed to imshow) | Alpha only via `imshow` |
| Redundant `isinstance(c, str)` filter in palette path | Removed (validated upstream) |
| White saturation with non-black-based cmaps (e.g. Reds, seismic) | Signal subtraction handles any cmap zero-value |
| `na_color` alpha compared via internal hex string | Uses `Color.get_alpha_as_float()` public API |

## Per-channel norm usage

```python
from matplotlib.colors import Normalize

sdata.pl.render_images(
    "morphology_focus",
    palette=["#0f73e6", "#a4a400", "#f300a5", "#008a00"],
    channel=["DAPI", "18S", "ATP1A1/CD45/E-Cadherin", "AlphaSMA/Vimentin"],
    norm=[Normalize(0, 5000), Normalize(0, 2000), Normalize(0, 8000), Normalize(0, 3000)],
).pl.show()
```

## Test plan

- [x] 5 existing baseline images updated (signal-based output differs from averaging)
- [x] 4 new visual tests with CI-generated baselines (5-channel compositing, per-channel norm)
- [x] 3 error tests for norm validation (length mismatch, empty list, tuple rejection)

Closes #416, #534. Addresses #370. Partially addresses #460 (per-channel norm, but not arbitrary preprocessing callables).

Note: #450 (PCA aggregation) is not addressed by this PR.

PR #471, opened 2025-06-04
fix image rendering (clipping warning)

Sonja-Stockhaus:
As discussed @timtreis, part of the code from #451 to do a first smaller pull request without the PCA approach


Sonja-Stockhaus:
Additional tests and the PCA aggregation for > 3 channels are in #451

PR #482, opened 2025-09-13
make ready for new spatialdata

melonora:
This PR ensures that spatialdata-plot works with the version of SpatialData on the zarrv3 branch.


melonora:
@timtreis @LucaMarconato merging this as it does not have really actual code changes in the codebase. It is more dealing with deprecations in spatialdata itself. We can decide whether we alter the spatialdata version on merging of the zarrv3 branch in SpatialData.

PR #492, opened 2025-09-23
`render_labels` and `render_shapes` now scan for and respect xyz_colors in adata.uns

timtreis:


src/spatialdata_plot/pl/utils.py
```diff
@@ -897,6 +928,139 @@ def _generate_base_categorial_color_mapping(
...
+        logger.debug(f"Color key '{color_key}' not found in table '{table_to_use}' uns.")
+        return None
+
+    # Extract colors and categories
+    stored_colors = adata.uns[color_key]
+    categories = color_source_vector.categories.tolist()
+
+    # Validate na_color format
+    if "#" not in str(na_color):
+        logger.warning("Expected `na_color` to be a hex color, converting...")
```
Sonja-Stockhaus, line 1024: hmm, do we really expect that the user passes a hex?

PR #511, opened 2025-11-15
When using `render_points()` with `datashader`, points can now be colored by `adata.obs`

timtreis:


grst (approved):
I confirm this solves my issue

PR #512, opened 2025-11-15
Safeguard against invalid user input in `render_shapes`

timtreis:


timtreis:
@josenimo could you rerun with this branch? What do you think of the error that is now returned?

PR #513, opened 2025-11-17
Polygons with holes are now rendered correctly

timtreis:


quentinblampey:
Hi @timtreis, thanks for the fix!
Sorry for the delay, I wanted to have a look tomorrow. I guess it'll look good, but I'll let you know if not

PR #515, opened 2025-11-18
Now raising an error if color or table doesn't exist

timtreis:
Now raising errors on missing input and if no table at all annotates the element.

In the process, I noticed we sometimes `warn` from our logger and sometimes from `warnings`, so I got rid of `warnings`. Then, I had to adjust the capturing of the logged messages, so I added that helper as well.


timtreis:
@LucaMarconato I assume this will have implications in our tutorials if we're blindly plotting elements that are not annotated.

PR #520, opened 2025-12-07
Multi-channel images now correctly normalised, even when uint16

timtreis:
Note: Image from `test_plot_extent_of_img_is_correct_after_spatial_query` changes because the image is now normalised to the value range of the cropped image instead of using the values from the uncropped image.


timtreis:
Hey @sophiamaedler @namsaraeva,

thanks for the bug report! Could you try out this PR?

Issue #462, opened 2025-05-13
Ability to change color of the border of labels


LucaMarconato:
Currently it doesn't seem possible to change the color of the border of labels.

For instance, reported by @MeyerBender, using 
```python
sdata_zoomed.pl.render_images(channel=["DAPI"]).pl.render_labels(outline_alpha=1, fill_alpha=0, color="white").pl.show()
```
shows a black border ([code here](https://haraldvohringer.com/spatialproteomics/notebooks/ExampleWorkflowSpatialdata.html))

<img width="534" alt="Image" src="https://github.com/user-attachments/assets/5a871a6a-d243-46ad-9dd7-3f9732301666" />

while having a white border like [here](https://haraldvohringer.com/spatialproteomics/notebooks/ExampleWorkflow.html) would be the desired output

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/f53171d3-435e-41df-bb5c-8d94c40669ca" />

MeyerBender:
Example code snippet:

```
import spatialdata as sd
import spatialdata_plot
from spatialdata.datasets import blobs

sdata = blobs()
sdata.pl.render_images(channel=0).pl.render_labels(outline_alpha=1, fill_alpha=0).pl.show()
```

It would be cool to have an argument like `contour_color` in there.

Issue #461, opened 2025-05-13
Color leak when `outline_alpha=1` and `full_alpha=0`


LucaMarconato:
Originally reported by @MeyerBender

> - normalization of channels: see differences between the two notebooks when plotting all channels but DAPI. I guess it is intentional that no normalization is performed by default, and it should be possible to normalize channels using the “norm” argument, but the documentation isn’t very clear on how to do this properly.

Using `outline_alpha=1` and `full_alpha=0` leads to plots like this one ([notebook here](https://haraldvohringer.com/spatialproteomics/notebooks/ExampleWorkflowSpatialdata.html)): 

<img width="520" alt="Image" src="https://github.com/user-attachments/assets/1383f662-2685-4969-9c99-76123ecf704e" />

Probably the erosion step misses some pixels.

MeyerBender:
Example code snippet:

```
import spatialdata as sd
import spatialdata_plot
from spatialdata.datasets import blobs

sdata = blobs()
sdata.pl.render_images(channel=0).pl.render_labels(outline_alpha=1, fill_alpha=0).pl.show()
```

<img width="623" alt="Image" src="https://github.com/user-attachments/assets/98e03a25-3e53-4865-8823-edc072207c26" />

For some reason this also creates two colorbars, not sure why.

MeyerBender:
Created a PR fixing this issue. Also, the two colorbars were the result of there being two images/labels in the `blobs` dataset (one normal and one multiscale). `sdata.pl.render_images('blobs_image', channel=0).pl.render_labels('blobs_labels', outline_alpha=1, fill_alpha=0).pl.show()` works as it should and only creates one colorbar.

PR #525, opened 2025-12-19
Feature/outline color for labels

MeyerBender:
This addresses two previous issues with rendering labels:
1. Users can now set the color of the outline with the `outline_color` argument, similar to how they could in `render_shapes()`, see #462 
2. Fixes color leak when plotting outlines only in `render_labels`, see #461 

Example code to see both changes in action:
```
import spatialdata as sd
import spatialdata_plot

sdata = blobs()
sdata.pl.render_images('blobs_image', channel=0).pl.render_labels('blobs_labels', outline_alpha=1, fill_alpha=0, outline_color='white').pl.show()
```


MeyerBender:
Hi @timtreis, this looks great, thanks a lot! As a sidenote, running `pl.render_labels('blobs_labels', outline_alpha=1)` without specifying a color sets a random color for each cell, which might be confusing. I agree with your comment in the other PR that using just one color would be more sensible in that case.

PR #549, opened 2026-03-12
`groups` now hides NAs unless `na_color` is specified

timtreis:
default: neither `groups` nor `na_color` is used -> NAs as grey

`groups` is used -> only shows the specified ones
`groups` is used and `na_color` is specified -> show NAs as `na_color` 

(also smuggled some cleanup in here)


timtreis:
@LucaMarconato FYI, this PR changes default behavior by aligning it across points, shapes and labels. Might affect notebooks

timtreis:
@LucaMarconato 

currently, when we're rendering labels without any further info, they're randomly colored because I felt that was pretty at that time. Now I'm wonderinf whether we should just render them in the default `na_color` but with a black outline?

It'd be less pretty by default but would make it very clear that there is no biology being displayed

PR #661, opened 2026-05-12
Raise clear error on disjoint instance IDs in render_shapes/labels/points

### LINKED ISSUE #603: render_shapes crashes with KeyError: None when table instance IDs don't match element
# `render_shapes` crashes with `KeyError: None` when table instance IDs don't match element

## Description

When calling `render_shapes("s", color="cat", table_name="t")` where the table correctly annotates element `"s"` (region key matches) but none of the table's `instance_id` values overlap with the element's actual instance IDs, the call crashes with a bare `KeyError: None` from deep inside pandas. The user gets a 15+ frame traceback that terminates in `element.loc[None, :]` with no indication of what went wrong or how to fix it.

Root cause: `join_spatialelement_table(..., how="inner")` returns an empty result. The downstream code passes `None` as an index to `element.loc`, which pandas raises as `KeyError: None`.

## Environment

```
spatialdata-plot: 0.3.4.dev (main, 5cfedc7)
spatialdata: 0.5.0
Python: 3.13
```

## Minimal Reproducible Example

```python
import matplotlib
matplotlib.use("Agg")
import numpy as np
import pandas as pd
import geopandas as gpd
import anndata as ad
import spatialdata as sd
from spatialdata.models import ShapesModel, TableModel
import spatialdata_plot
from shapely.geometry import Point

shapes = ShapesModel.parse(gpd.GeoDataFrame({
    "geometry": [Point(5, 5), Point(15, 5), Point(25, 5)],
    "radius": [2.0] * 3,
}))

obs = pd.DataFrame({
    "instance_id": [99, 100, 101],  # element has IDs 0, 1, 2 — no overlap
    "region": pd.Categorical(["s"] * 3),
    "cat": pd.Categorical(["A", "B", "C"]),
})
obs.index = obs.index.astype(str)

table = TableModel.parse(
    ad.AnnData(X=np.zeros((3, 1)), obs=obs),
    region=["s"], region_key="region", instance_key="instance_id",
)
sdata = sd.SpatialData(shapes={"s": shapes}, tables={"t": table})

fig, ax = plt.subplots()
sdata.pl.render_shapes("s", color="cat", table_name="t").pl.show(ax=ax)
```

## Actual Output

```
KeyError: None
```

Full traceback ends at `element.loc[None, :]` buried in spatialdata internals (`_get_masked_element`). No indication that the instance ID sets are disjoint.

## Expected Output

Either:
- All shapes rendered in `na_color` with a `UserWarning`: `"No instance IDs in table 't' overlap with element 's'. Shapes will be rendered with na_color."`
- Or a clear `ValueError`: `"No instance IDs overlap between table 't' (instance_key='instance_id') and element 's'. Check that the table's instance_id column matches the element's index."`

## Fix Sketch

In `_render_shapes` (and analogously `_render_labels`, `_render_points`), before calling `join_spatialelement_table`, check whether the intersection of table instance IDs and element instance IDs is empty. If so, warn and skip coloring (fall back to `na_color`) or raise a descriptive `ValueError`.

```python
element_ids = set(element.index)
table_ids = set(table.obs[instance_key])
if element_ids.isdisjoint(table_ids):
    raise ValueError(
        f"No instance IDs overlap between table '{table_name}' "
        f"(instance_key='{instance_key}') and element '{element_name}'. "
        f"Check that the table's instance_id column matches the element's index."
    )
```

Note: the root cause lives upstream in `spatialdata._get_masked_element`, but a guard in `spatialdata-plot` provides immediate user-facing relief without requiring a coordinated upstream fix.

---
**Triage tier**: Tier 2

SAY-5:
Closes #603.

When a table annotates the element (region key matches) but no `instance_id` values overlap, the call crashed with a bare `KeyError: None` from deep inside upstream `spatialdata`'s join path. The traceback gave no indication that the instance ID sets were disjoint.

Added an explicit disjoint check before `join_spatialelement_table`: when at least one table row claims to annotate the element yet the `instance_key` values share nothing with the element's index, raise a `ValueError` naming the table, element, instance key, and the fix. Regression test in `test_render_shapes.py`.

### DIFF
--- a/src/spatialdata_plot/pl/render.py
+++ b/src/spatialdata_plot/pl/render.py
@@ -311,6 +311,23 @@ def _add_legend_and_colorbar(
     )
 
 
+def _check_instance_ids_overlap(
+    sdata: sd.SpatialData,
+    table_name: str,
+    element_name: str,
+    element_index: abc.Iterable[Any],
+) -> None:
+    """Raise a clear error when a table annotates an element but no instance IDs overlap (#603)."""
+    _, region_key, instance_key = get_table_keys(sdata[table_name])
+    annotating = sdata[table_name].obs[sdata[table_name].obs[region_key].isin([element_name])]
+    if len(annotating) > 0 and set(annotating[instance_key]).isdisjoint(set(element_index)):
+        raise ValueError(
+            f"No instance IDs overlap between table '{table_name}' (instance_key='{instance_key}') "
+            f"and element '{element_name}'. Check that the table's '{instance_key}' column matches the "
+            f"element's index."
+        )
+
+
 def _render_shapes(
     sdata: sd.SpatialData,
     render_params: ShapesRenderParams,
@@ -336,6 +353,8 @@ def _render_shapes(
         table = None
         shapes = sdata_filt[element]
     else:
+        _check_instance_ids_overlap(sdata_filt, table_name, element, sdata_filt[element].index)
+
         # Workaround for upstream spatialdata bug (scverse/spatialdata#1099):
         # join_spatialelement_table calls table.obs.reset_index() which fails when
         # the obs index name matches an existing column (e.g. "EntityID" in Merfish data).
@@ -742,6 +761,9 @@ def _render_points(
 
     added_color_from_table = False
     if col_for_color is not None and col_for_color not in points.columns:
+        if table_name is not None:
+            # guard against disjoint instance IDs (#603) for a clearer error than KeyError: None
+            _check_instance_ids_overlap(sdata_filt, table_name, element, points.index)
         color_values = get_values(
             value_key=col_for_color,
             sdata=sdata_filt,
@@ -1651,6 +1673,7 @@ def _render_labels(
         instance_id = np.unique(label)
         table = None
     else:
+        _check_instance_ids_overlap(sdata_filt, table_name, element, np.unique(label.values))
         _, region_key, instance_key = get_table_keys(sdata[table_name])
         table = sdata[table_name][sdata[table_name].obs[region_key].isin([element])]
 
--- a/src/spatialdata_plot/pl/render.py
+++ b/src/spatialdata_plot/pl/render.py
@@ -317,7 +317,7 @@ def _check_instance_ids_overlap(
     element_name: str,
     element_index: abc.Iterable[Any],
 ) -> None:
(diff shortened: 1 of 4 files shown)


src/spatialdata_plot/pl/render.py
```diff
@@ -345,6 +345,18 @@ def _render_shapes(
         if _saved_index_name is not None and _saved_index_name in _obs.columns:
             _obs.index.name = None
 
+        # Guard against a disjoint instance_key (#603): when no table rows annotate any element
```
timtreis: Claude tends to be overly verbose, please remove this comment

tests/pl/test_render_shapes.py
```diff
@@ -1323,6 +1323,39 @@ def test_render_shapes_color_with_conflicting_index_name():
     sdata.pl.render_shapes("shapes", color="cell_type", table_name="table").pl.show()
 
 
+def test_render_shapes_disjoint_instance_ids_clear_error():
+    # Regression test for #603: when a table annotates the element (region key matches) but no
```
timtreis: This is also overly verbose, reduce to one-liner

tests/pl/test_render_shapes.py
```diff
@@ -1323,6 +1323,39 @@ def test_render_shapes_color_with_conflicting_index_name():
...
+    # instance_id values overlap, the call used to crash with a bare `KeyError: None` from deep
+    # inside spatialdata's join path.  Replace with a clear, actionable ValueError.
+    from shapely.geometry import Point
```
timtreis: Don't reimport inside the test

src/spatialdata_plot/pl/render.py
```diff
@@ -345,6 +345,18 @@ def _render_shapes(
...
+        # row, the upstream join crashes with `KeyError: None` from deep inside pandas.  Raise
+        # a clear, actionable ValueError before invoking the join.
+        _, _region_key, _instance_key = get_table_keys(sdata[table_name])
```
timtreis: Operate on `sdata_filt` instead of `sdata` here

SAY-5:
Switched the overlap check to operate on sdata_filt and dropped the verbose comments. Rebased on main so there are no conflicts. New regression tests pass locally.

PR #676, opened 2026-05-20
Add method='datashader' to render_images for sparse images

### LINKED ISSUE #449: Feature: datashader also for images
As shown in this thread here: https://github.com/scverse/spatialdata-notebooks/pull/138, when using `imshow()` on very sparse matrices, the resulting plot can be misleading (the same problem appears also when plotting data using in `napari`).

A solution could be to enable a `method='datashader'` approach also for `render_images`, so that a 2D histogram is computed and the interpolation artifacts can be better controlled.

timtreis:
## Summary
- Closes #449. Adds `method='datashader'` and `ds_reduction` kwargs to `render_images()` so very sparse images (mostly zeros, rare non-zero pixels) survive the downsample step instead of being averaged to near-black by the default mean-aggregating rasterization + `imshow` interpolation. Mirrors the existing `render_points`/`render_shapes` API.
- Routes the rasterization step through `datashader.Canvas.raster` with a configurable downsample method (default `'max'`); forces `imshow(interpolation='nearest')` on that path so the chosen reduction isn't re-smoothed at display time. Geometry + spatial transformations are inherited from `spatialdata.rasterize` so the output integrates cleanly with the existing transform plumbing.
- Centralizes `_DsReduction` (previously inline-duplicated across 5 sites) into `render_params.py`, alongside a new `_ImageDsReduction` for the image-only valid set (`mode`, `first`, `last` added; `sum`, `any`, `count` removed since they're not valid `Canvas.raster` `downsample_method` values).

## API addition (purely additive, defaults preserve current behavior)

```python
render_images(
    ...,
    method: Literal["matplotlib", "datashader"] | None = None,
    ds_reduction: Literal["max", "min", "mean", "mode", "first", "last", "var", "std"] | None = None,
)
```

`method=None` (default) is byte-for-byte identical to the existing path.


timtreis:
FYI @LucaMarconato

PR #684, opened 2026-05-21
Add interactive annotation ability via `.pl.annotate()`

timtreis:
Adds `sdata.pl.annotate()`, a terminal step on a render chain that hands the rasterised RGB to [`anybioimage`](https://github.com/maartenpaul/anybioimage)'s `BioImageViewer`, then commits drawn shapes back to `sdata.shapes` as a `ShapesModel` in the chosen CS with `Identity()`.

- New optional extra: `pip install 'spatialdata-plot[interactive]'`.
- Works over SSH; rectangle / polygon / point tools.
- `_copy()` carries a `_source_sdata` ref so writeback lands on the user's `sdata`, not the chain copy.


LucaMarconato:
Hi, really useful feature! What's the relationship to this https://github.com/scverse/2026_04_hackathon_padua/issues/22?

timtreis:
> Hi, really useful feature! What's the relationship to this [scverse/2026_04_hackathon_padua#22](https://github.com/scverse/2026_04_hackathon_padua/issues/22)?

I've now refactored the feature to use `anybioimage` instead of directly shipping an equivalent widget inside the package, hopefully it'll keep stable 🤞

PR #686, opened 2026-05-28
Add scanpy-style multi-panel `color=[...]` for render_shapes and render_labels

### LINKED ISSUE #611: Feature request: support color=["gene1","gene2"] for multi-panel and multi-channel spatial plots
# Feature request: support `color=["gene1","gene2"]` for multi-panel and multi-channel spatial plots

## Summary

All `render_*` functions accept only a single `color=` value. There is no way to generate multi-panel plots automatically from a list of columns (like `sc.pl.umap(color=["gene1","gene2"])` in scanpy), and no way to create multi-gene composite overlays (like Xenium Explorer's multi-stain panel). This is consistently the most-requested missing feature across the issue tracker.

## Environment

```
spatialdata-plot: 0.3.4.dev (main, 5cfedc7)
spatialdata: 0.5.0
Python: 3.13
```

## Evidence from Issues

- **#321**: "It would be nice for the parameter `color` to support gene list" — users expect scanpy-style `color=['gene1','gene2']` to produce one panel per gene.
- **#534**: Users need per-marker spatial panels (Xenium Explorer-style) but must write extensive boilerplate.
- **#450**: Requests per-channel-norm plots as separate panels when compositing is not possible (N≥4 channels).
- **#269**: Consequence of no multi-panel support: `ncols` parameter in `show()` has no documented use case.

## Current Workaround

Verbose boilerplate required for just 3 genes:

```python
import spatialdata as sd
import spatialdata_plot

sdata = sd.datasets.blobs()
genes = ["feature_0", "feature_1", "feature_2"]

fig, axes = plt.subplots(1, len(genes), figsize=(15, 5))
for ax, gene in zip(axes, genes):
    sdata.pl.render_shapes("blobs_circles", color=gene).pl.show(ax=ax)
    ax.set_title(gene)
plt.tight_layout()
```

## Desired API

Consistent with scanpy's `color=` list convention:

```python
import spatialdata as sd
import spatialdata_plot

sdata = sd.datasets.blobs()

# Multi-panel: one subplot per gene
sdata.pl.render_shapes("blobs_circles", color=["feature_0", "feature_1", "feature_2"]).pl.show(ncols=3)

# Multi-channel composite: overlay two annotation columns
sdata.pl.render_labels("blobs_labels", color=["cat1", "cat2"]).pl.show()
```

## Proposed Behavior

- `color=["gene1","gene2"]` auto-creates a subplot grid (respecting `ncols=`) with one panel per gene, matching `sc.pl.umap(color=[...])` behavior.
- `ncols` in `show()` controls how many panels fit per row.
- Each panel shares the same spatial extent and coordinate system.
- The existing `ncols` parameter in `show()` — currently undocumented as functional — becomes the natural row-width control for this feature.

## Related

Issues #321, #450, #534. The `ncols` parameter in `show()` is currently undocumented as functional.

**Labels**: enhancement, priority: high

---
**Triage tier**: Tier 2

timtreis:
## Summary

Closes #611. `render_shapes` and `render_labels` now accept a **list of column/key names** for `color`, producing one panel per key — the scanpy `color=[...]` convention.

```python
sdata.pl.render_shapes("blobs_circles", color=["gene1", "gene2", "gene3"]).pl.show(ncols=3)
```

- A string list is expanded **at render time** into one set of render params per key, reusing the existing single-color validation unchanged; each entry is tagged with a new `panel_key`.
- `show()` lays out one panel per key; scalar-colored render calls (e.g. a background image) are drawn into **every** panel as a shared layer.
- `show(ncols=...)` controls the grid width; each panel is titled by its key.

A `list[str]` is statically distinct from `ColorLike`'s `list[float]`, so the type surface stays clean (`color: ColorLike | list[str] | None`).

## Behavior & constraints

- **Additive / backward compatible**: scalar, RGBA, and column `color` calls are unchanged. A length-1 list normalizes to a scalar color.
- Multi-panel color requires a **single coordinate system** (clear error otherwise, naming the systems).
- Only **one** `render_*` call per figure may pass a list (others must be scalar).
- Duplicate keys are rejected; invalid keys are aggregated into one error **before** any drawing.
- `groups` / `palette` / `cmap` / `norm` are broadcast to every panel; each continuous panel auto-scales independently; colorbars/legends are per-panel.

## Drive-by fix

`_copy()` now shallow-copies `plotting_tree`, so appending a render step to one chain no longer mutates a sibling chain branched off the same object (this previously also let branched chains bleed layers into each other).


LucaMarconato:
CCing @melonora, wasn't this feature available on a earlier version and then got refactored out because it was making linking parameters to color challenging?

But I think having all the elements in `color=` share the other parameters would be a good way to have this back.

timtreis:
Yeah, we removed it for type checking "elegance" but with the new `ColorLike` class it's cleaner now. And it's annoying for people to not have this ability in the native package.

The original plan was to upstream that ability to Squidpy, but there are so many plots you can only do with the non-wrapped `sdata-plot` that I figured it'd be easier to just allow it here again so people don't have to go upstream for this one specific feature

PR #705, opened 2026-06-08
Add `measure_obs`: persist per-cell centroid/area/equivalent diameter into the annotating table

timtreis:
## What
Public `measure_obs` utility — computes per-cell **centroid**, **area** and **equivalent diameter** for a shapes or 2D-labels element and writes them into the annotating `AnnData` table (squidpy-style):

- centroid → `obsm["spatial"]` · area → `obs["area"]` · equiv. diameter → `obs["equivalent_diameter"]`

Stored in the element's **intrinsic** units. Labels area = pixel count; shapes area = `geometry.area` (`pi*r**2` for circles).

```python
from spatialdata_plot.pl import measure_obs

measure_obs(sdata, "cells")        # in place
measure_obs(sdata, inplace=False)  # returns a copy
```

## Why
Persist centroids/area once so renders and downstream tools (squidpy) reuse them instead of recomputing. `obsm["spatial"]` is the canonical, coords-only home; area belongs in `obs`.

## How
- **Labels**: streaming bincount aggregator, block-by-block (one chunk + `O(n_labels)` accumulators) — out-of-core, scales to Xenium-size masks; area is a free by-product.
- **Shapes**: shapely vectorized centroid/area; circles (`Point`+`radius`) use `pi*r**2`.
- Compute-and-write (overwrites); `centroids=False` keeps an existing `obsm["spatial"]`. Needs an annotating table. `inplace` follows the scanpy convention.

## Scope
Utility only — wiring `as_points` rendering through these measurements is a follow-up.

Tested in `tests/pl/test_utils.py::TestMeasureObs`; performance benchmarks in the comment below.

### DIFF
--- a/src/spatialdata_plot/pl/__init__.py
+++ b/src/spatialdata_plot/pl/__init__.py
@@ -1,8 +1,10 @@
 from ._palette import make_palette, make_palette_from_data
 from .basic import PlotAccessor
+from .utils import measure_obs
 
 __all__ = [
     "PlotAccessor",
     "make_palette",
     "make_palette_from_data",
+    "measure_obs",
 ]
--- a/src/spatialdata_plot/pl/__init__.py
+++ b/src/spatialdata_plot/pl/__init__.py
@@ -1,10 +1,8 @@
 from ._palette import make_palette, make_palette_from_data
 from .basic import PlotAccessor
-from .utils import measure_obs
 
 __all__ = [
     "PlotAccessor",
     "make_palette",
     "make_palette_from_data",
-    "measure_obs",
 ]
--- a/src/spatialdata_plot/pl/utils.py
+++ b/src/spatialdata_plot/pl/utils.py
@@ -26,6 +26,7 @@ import shapely
 import spatialdata as sd
 from anndata import AnnData
 from cycler import Cycler, cycler
+from dask.array.core import slices_from_chunks
 from datashader.core import Canvas
 from geopandas import GeoDataFrame
 from matplotlib import colors, patheffects, rcParams
@@ -59,6 +60,7 @@ from skimage.morphology import erosion, footprint_rectangle
 from skimage.util import map_array
 from spatialdata import (
     SpatialData,
+    deepcopy as sd_deepcopy,
     get_element_annotators,
     get_extent,
     get_values,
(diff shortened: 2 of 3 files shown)


timtreis:
## Performance (real data)

`get_centroids`'s labels backend loops per raster slice (`da.unique` ×2 per row/column); the streaming bincount replaces it. Centroids byte-identical (~1e-14).

**Real cell-segmentation masks**

| dataset | cells | `get_centroids` | bincount | speedup |
|---|---|---|---|---|
| mibitof (1024²) | ~1.1–1.5k | ~33 s | ~31 ms | ~1000× |
| small_test_region (800²) | 2,622 | 39.7 s | 32 ms | **1257×** |

**Scale / out-of-core** — real nucleus mask tiled to **671,232 cells / 164 M px**, measured end-to-end (read from disk + persist) in **5.6 s** at flat ~590 MB peak (4× the pixels → same memory; accumulators 16 MB). Real Visium HD: **5.48 M shapes in 4.4 s**.

A real-data run also surfaced a circle-area bug (shapely `Point.area == 0`), fixed here.

LucaMarconato:
A comment on the latest message. [With this PR](https://github.com/scverse/spatialdata/commit/6ee10136f3b1f95386af890835b029342359f56f) (see in particular the text below), `get_centroids()` are optimized.

```
* perf: vectorize label centroid computation, 30x speedup

Replace per-slice O(H+W) approach (512 dask compute() calls for a 256×256
array) with a single array materialization + np.bincount O(n_pixels) pass.

This speeds up get_centroids() on labels from ~1.5s to ~50ms, cutting
to_circles(labels) from ~1.6s to ~53ms. Affects test_validation dataloader
variants (~2.5s → ~0.2s each, saving ~9s), test_labels_2d_to_circles, and
any production call to get_centroids or to_circles on label arrays.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
```

Two questions:
- the above is available from `spatialdata==0.7.3`, are you using it?
- it seems that your implementation is even faster. If so, could you please upstream it?

timtreis:
Not sure if I had the version from sdata 0.7.3, but I'll prototype the plotting speedups with this one here and if I like the UX, I'll upstream 👌 If it doesn't hold up, I'll just kick it out again before the next release

timtreis:
I think the major benefit of my approach is that I'm getting the area for next to no extra cost as well which is also super useful for plotting + computations

PR #742, opened 2026-06-22
fix: align rasterized artists with get_extent 

### LINKED ISSUE #216: pixel shift in `render_images`
Hello,

When plotting an image, it seems that there is a 0.5-pixel shift. This is not visible for large images, but becomes very visible if we zoom in.

### Minimum reproducible example

```python
import numpy as np
import spatialdata_plot
import spatialdata
from spatialdata.models import Image2DModel
from spatial_image import SpatialImage

image = SpatialImage(np.zeros((1, 4, 8)), dims=("c", "y", "x"))
image = Image2DModel.parse(image)

sdata = spatialdata.SpatialData(images={"image": image})

sdata.pl.render_images().pl.show()
```

It should render this image (we see that the image is not reaching `(8, 4)`, but `(7.5, 3.5)` instead):
image

Maybe I did something wrong with the x-coords or y-coords? Should we update the coords or they should be fine by default?

Sorry for such a minor issue...

timtreis:
Closes #216.

## Problem
Rasterized artists (images, labels, datashader output) used matplotlib's default **pixel-center** imshow extent `(-0.5, W-0.5, ...)`, while axis limits (`get_extent`), the element affine, and point/shape overlays use the **pixel-edge** convention. So every image sat half a pixel off — and the affine amplifies it (a `Scale(1000)` image shifted by 500 world units).

## Fix
Use pixel-edge extent `(0, W, H, 0)` so each artist's data box is the `[0, shape]` box `get_extent` transforms — they coincide under any affine.
- `_datashader.py` `_ax_show_and_transform` — shared by images + datashader.
- `render.py` `_render_labels` — labels bypass that helper via a direct `imshow(origin="lower")`, so they get an explicit `extent=(0, W, 0, H)`.

This also removes the residual datashader-points offset relative to the matplotlib backend.

## Why pixel-edge
It's spatialdata's own convention: `get_extent` reports an image as occupying `[0, shape]`, and `get_centroids` puts pixel `[0,0]`'s centroid at `(0.5, 0.5)`. The fix matches the data model — a `get_centroids` overlay now lands dead-center on its label pixel instead of on the corner.


timtreis:
FYI @LucaMarconato 

ignored until now because I didn't think it'd matter much on real data, but I guess good that it's eventually fixed

PR #745, opened 2026-06-24
fix: `pl.show()` now displays in a plain Python REPL

### LINKED ISSUE #68: `pl.show()` still not working (?) in interactive mode
I am using the latest code from main. If I run this code as a script, the plot is shown. If I set a breakpoint (I am using PyCharm) in the line with `pl.render_points()...`, and then run that line in the Python interactive console, then no plot is shown and I need to call `plt.show()` explicitly. I think the bug was once fixed, maybe it slipped in again. 

```
from spatialdata import SpatialData
from spatialdata.models import PointsModel, ShapesModel
from shapely import polygons, linearrings
from geopandas import GeoDataFrame
import pandas as pd
import geopandas as gpd
import numpy as np

def _make_points():
    """Helper function to make a Points element."""
    coordinates = np.array([[10, 10], [20, 20], [20, 30]], dtype=float)
    return PointsModel.parse(
        coordinates, annotation=pd.DataFrame({"genes": np.repeat("a", len(coordinates))}), feature_key="genes"
    )


def _make_squares() -> polygons:
    centroid_coordinates = np.array([[10, 10], [10, 80], [80, 20], [70, 60]])
    half_width = 6
    linear_rings = []
    for centroid in centroid_coordinates:
        min_coords = centroid - half_width
        max_coords = centroid + half_width

        linear_rings.append(
            linearrings(
                [
                    [min_coords[0], min_coords[1]],
                    [min_coords[0], max_coords[1]],
                    [max_coords[0], max_coords[1]],
                    [max_coords[0], min_coords[1]],
                ]
            )
        )
    s = polygons(linear_rings)
    polygon_series = gpd.GeoSeries(s)
    cell_polygon_table = gpd.GeoDataFrame(geometry=polygon_series)
    sd_polygons = ShapesModel.parse(cell_polygon_table)
    return sd_polygons


def _make_circles() -> GeoDataFrame:
    centroid_coordinates = np.array([[10, 10], [10, 80], [80, 20], [70, 60]])
    radius = 10
    circles = ShapesModel.parse(centroid_coordinates, geometry=0, radius=radius)
    return circles

points = _make_points()
squares = _make_squares()
circles = _make_circles()
import spatialdata_plot
sdata = SpatialData(points={'p': points}, shapes={'s': squares, 'c': circles})
sdata.pl.render_points().pl.render_shapes().pl.show()
# import matplotlib.pyplot as plt
# plt.show()
```

timtreis:
## Problem

Fixes #68. `pl.show()` did not display the figure in a plain Python REPL (PyCharm console, pasted REPL, `python -i`); users had to call `plt.show()` manually. Most recently re-reported [here](https://github.com/scverse/spatialdata-plot/issues/68#issuecomment-4787594596).

## Root cause

The `show=None` default keyed off `not hasattr(sys, "ps1")`. A plain Python REPL sets `sys.ps1` but leaves matplotlib **non-interactive** (unlike Jupyter/IPython), so `plt.show()` was suppressed and nothing rendered.

## Fix

Drop the `sys.ps1` check and decide purely on `matplotlib.is_interactive()`:

- interactive backends (Jupyter, IPython, `plt.ion()`) auto-render → no `plt.show()`;
- scripts and plain REPLs are non-interactive → explicit `plt.show()`.

The new condition is strictly more permissive than the old one, so it never suppresses a `show()` the old code emitted — it only adds the missing REPL case. Removed the now-unused `import sys`.

## Tests

Added `test_show_default_keys_off_is_interactive` (parametrized over the interactive flag, `sys.ps1` set in both cases to prove the decision ignores it). Verified to fail on the old logic and pass on the fix.

### DIFF
--- a/src/spatialdata_plot/pl/basic.py
+++ b/src/spatialdata_plot/pl/basic.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import sys
 import warnings
 from collections import OrderedDict
 from collections.abc import Callable, Sequence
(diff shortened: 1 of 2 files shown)


timtreis:
@LucaMarconato did this issue also affect you?

LucaMarconato:
Amazing Tim! Tried it out and the bug is gone 😊

Issue #606, opened 2026-04-24
render_labels crashes with cryptic skimage TypeError when label array has float dtype


timtreis:
# `render_labels` crashes with cryptic skimage `TypeError` when label array has float dtype

## Description

`Labels2DModel.parse` accepts `float32` arrays without error — this is common output from ML segmentation pipelines. When `render_labels` is subsequently called on such an element, it crashes deep in `_map_color_seg` with `TypeError: The dtype of an array to be remapped should be integer.` from skimage internals. The traceback contains no mention of the element name, the offending dtype, or how to fix the problem.

## Environment

```
spatialdata-plot: 0.3.4.dev (main, 5cfedc7)
spatialdata: 0.5.0
Python: 3.13
```

## Minimal Reproducible Example

```python
import matplotlib
matplotlib.use("Agg")
import matplotlib.pyplot as plt
import numpy as np
import spatialdata as sd
from spatialdata.models import Labels2DModel
import spatialdata_plot

# Labels from ML model output — commonly float32
labels = np.zeros((20, 20), dtype=np.float32)
labels[3:8, 3:8] = 1.0
labels[12:17, 12:17] = 2.0

sdata = sd.SpatialData(labels={"lbl": Labels2DModel.parse(labels, dims=["y", "x"])})
fig, ax = plt.subplots()
sdata.pl.render_labels("lbl").pl.show(ax=ax)
```

## Actual Output

```
TypeError: The dtype of an array to be remapped should be integer.
```

The full traceback runs through `skimage/util/_map_array.py` with no spatialdata-plot frame visible. No mention of the element name, its dtype, or what to do about it.

## Expected Output

A clear, actionable error at the entry point to `_render_labels`:

```
ValueError: Label element 'lbl' has dtype float32. Label arrays must use an integer
dtype (e.g. int32 or uint16). Cast before plotting:
    sdata['lbl'] = sdata['lbl'].astype(np.int32)
```

## Fix Sketch

In `_render_labels`, after the label array is loaded and squeezed, add a dtype guard before any downstream processing:

```python
if np.issubdtype(label.values.dtype, np.floating):
    raise ValueError(
        f"Label element '{element_name}' has dtype {label.values.dtype}. "
        "Label arrays must use an integer dtype (e.g. int32 or uint16). "
        "Cast before plotting:\n"
        f"    import numpy as np\n"
        f"    sdata['{element_name}'] = sdata['{element_name}'].astype(np.int32)"
    )
```

[...]

PR #751, opened 2026-07-10
fix(labels): restore render float-dtype guard for spatialdata <0.8

timtreis:
## Problem

#758 (merged) removed `_render_labels`' float-dtype guard, assuming spatialdata always rejects float labels at the model boundary. That only holds on **spatialdata >= 0.8**. The library still declares **`spatialdata >= 0.3`**, and on spatialdata < 0.8 `Labels2DModel.parse` *accepts* float labels — so with the guard gone, `render_labels` again crashes deep in skimage with a cryptic `TypeError` (the #606 regression). This PR corrects that.

## Changes
- **render.py** — restore the render-time float-dtype `ValueError` guard (safety net for spatialdata < 0.8; >= 0.8 rejects earlier at the model boundary).
- **test_render_labels.py** — re-add `test_render_labels_rejects_float_dtype`, now version-agnostic: asserts an `"integer dtype"` error at whichever layer enforces it (spatialdata on >= 0.8; our guard on < 0.8).
- **test_render_shapes.py** — `sort_index()` the queried shapes before rendering the random-shuffling test, so the overlapping circles' draw order (and baseline) is stable across spatialdata's 0.8 query-order change; baseline regenerated.

`measure_obs` float coverage stays as the version-independent `_stream_label_centroid_stats` unit test added in #758.


timtreis:
Blocked by https://github.com/scverse/spatialdata/pull/1150 so we can remove the local `measure_obs` again

PR #765, opened 2026-08-11
feat(show): crop kwarg for on-the-fly bounding-box rendering

### LINKED ISSUE #764: Crop rendering to a coordinate-system bounding box from show()
There is no render-level crop. To restrict a plot to a region, callers pre-query with `sdata.query.bounding_box(...)` and render the subset, or render everything and clip with `ax.set_xlim`/`ax.set_ylim` afterward. The first is verbose; the second rasterizes the full image before clipping, which is wasteful at Visium HD scale (100k x 100k H&E).

Request: a `crop` kwarg on `show()` (or `render_*`) taking a bounding box in the target coordinate system, e.g. `crop=(x0, x1, y0, y1)`, that limits both the query extent and the rendered canvas so large images are not fully rasterized.

- Per-coordinate-system when multiple systems are shown.
- Squidpy exposes this today as `spatial_scatter(..., crop_coord=...)`; a native crop would let delegating wrappers drop their manual clip path.

Version: 0.4.1.

timtreis:
Closes #764.

Adds `show(crop_coord=(xmin, xmax, ymin, ymax))` to restrict a plot to a bounding box in the rendered coordinate system's units. Limits both the data drawn and the rasterized canvas, so large elements are not fully materialized before being clipped. Named `crop_coord` to match `sc.pl.spatial` / `sq.pl.spatial_scatter` (same 4-tuple order).

```python
sdata.pl.render_images("he").pl.show(crop_coord=(x0, x1, y0, y1))
```

## Per element type

- points / shapes: predicate subset before drawing (dask predicate for points; radius-aware bounds for circles, bbox-intersect for polygons, per-row so mixed circle/polygon frames are correct). Colour domain is taken from the full element, so colours and colourbar match the uncropped plot (resolved only on the matplotlib path; transfunc-aware).
- images (matplotlib and datashader): rasterize only the crop window via `rasterize()`'s target bbox. The full image is never materialized and the window keeps full figure resolution. Datashader images aggregate only the sliced source, preserving the configured reduction. Contrast auto-scales over the window (`vmin`/`vmax`/`norm` to override).
- labels: rasterized to the window like images (multiscale picks the level for the window). Stable for numeric, pandas-Categorical, and explicit-palette columns; a plain-string column with the default palette falls back to full render + clip (windowing would otherwise reshuffle colours as categories drop out of the window). Empty-window guard.

Axis limits are set to the exact box (bypasses the extent-union merge, honours the inverted y-axis, ignores `pad_extent`). Restricted to a single coordinate system (validated).

Rotation/shear falls back to full render + clip (a box does not map to a box). Multi-coordinate-system crop with a per-CS box is not implemented.

## Benchmarks

Prototype scripts, 3% window, image carrying a Scale+Translation. Two things measured: matplotlib image rasterization time vs image size, and datashader-image peak memory.

Matplotlib image, window rasterize vs full-render-then-clip (wall time):

| image | full + clip | window | speedup |
|-------|-------------|--------|---------|
| 6000² | 57 ms | 16 ms | 3.6× |
| 12000² | 173 ms | 16 ms | 10.8× |
| 20000² | 436 ms | 17 ms | 25.7× |

Window cost is flat (~16 ms) because only the window's source pixels are read; the full path grows with image size and would OOM at Visium HD scale.

Datashader image (8000² sparse): peak memory 258 MB → 19 MB, since the full image is no longer passed through `image.compute()`.

## Tests

`tests/pl/test_show.py`, `tests/pl/test_utils.py`: bbox helpers (negative-scale re-sort, rotation → None, radius-aware + mixed circle/polygon), points/shapes subset, full-element colour domain, transfunc match, datashader window autoscale, multiscale level selection, image and datashader-image window rasterization, labels placement / empty-window / no-colour-reshuffle. Full non-visual `pl` suite: 547 passed.


timtreis:
Follow-up `5b66f8d` addresses a max-effort review:

**API naming** — renamed the kwarg `crop` → **`crop_coord`** to match `sc.pl.spatial` / `sq.pl.spatial_scatter` (the 4-tuple order already matched; internals still map to `bounding_box_query`'s `min/max_coordinate` at the adapter layer).

**Correctness**
- Norm handling is now **non-mutating**: the full-element range is captured and applied **only on the matplotlib backend**. This fixes (a) the datashader leak — it now autoscales over the window (correct for `sum`/`count`, which can't match uncropped), and (b) the pre-`transfunc` pin — the range is taken over the transfunc'd full vector.
- `_bbox_mask_shapes` is now **radius-aware per row**, so a frame mixing circles and polygons no longer drops edge circles.
- **Multiscale** images select the pyramid level for the crop **window's** required resolution, not the full image — a Visium-HD crop no longer upsamples a coarse level.

**Scope/docs** — `crop_coord` documents that labels and `method='datashader'` images are clip-only (windowed labels need full-element categorical-palette handling; deferred). Axis-limit block simplified to read `crop_coord` directly.

**Tests added** (the review's gap): transfunc-matches-uncropped, datashader-autoscales-over-window, multiscale-selects-finer-level, mixed circle+polygon mask. Full non-visual suite: 361 passed.

Remaining follow-ups: windowed labels, datashader-**image** window path (materializes the full image in the `Canvas.raster` step — larger change), multi-CS per-box crop, rotated/sheared fast path.

timtreis:
Follow-up `906b811` — **datashader-image crop is now windowed too** (was the last deferred item).

`method='datashader'` images previously ignored `crop_coord` (full render + axis-clip). Now the aggregation is restricted to the window: `rasterize()` supplies the correct window coords/transform, only the sliced source is materialized, and the configurable reduction (`max`/`sum`/…) is preserved by aggregating the sliced source in pixel space and index-assigning onto the rasterized grid. The full image is never read.

Prototype (`profiling/bbox_ds_image_crop_prototype.py`, 8k² sparse image, Scale+Translation, 5% window): peak memory **258 → 19 MB** and O(window) instead of O(image) — so it scales to Visium HD where the old path would OOM on `image.compute()`. Placement verified against the window extent; the sparse peak survives the reduction; a value far outside the window is correctly excluded. Falls back to full render for rotation/shear or an empty window.

Regression test added; full non-visual suite still green (227 image/utils + 12 crop).

Only **labels** remain clip-only under crop now (windowing them needs full-element categorical-palette handling) — documented as a follow-up.

timtreis:
Follow-up `c84f76a` — **labels crop is now windowed too**, so every element type honours `crop_coord`'s fast path (the last deferred item).

`_render_labels` threads `crop` into the same crop-aware `_rasterize_if_necessary` / `_multiscale_to_spatial_image` used by images: the label raster is rasterized to the window (full array never materialized; multiscale picks the level for the window's resolution).

**Colour stability** (the reshuffle concern from the review): windowing drops off-window instances before colours resolve. That's stable for numeric columns, **pandas-Categorical** columns (fixed levels) and explicit palettes. A **plain-string** column with the default palette *would* reshuffle (as whole categories drop out of the window their sorted positions shift), so that specific case **falls back to full render + clip** — colours always match the uncropped plot; use a categorical dtype or an explicit palette to keep the windowed fast path. An empty-window guard returns cleanly when the box covers no labels.

Tests: placement + empty-window; no-colour-reshuffle for categorical (windowed) and string (fallback). Full non-visual suite: **390 passed**.

Every element type now has the crop fast path: points/shapes (predicate subset), images + datashader-images (window rasterize), labels (window rasterize with the string-palette fallback). Only rotation/shear and multi-CS per-box remain as documented follow-ups.

